Issue #5, opened 2020-04-01
Cached credentials are not checked when in legacy mode


danmx:
After following kube config the [conversion script](https://github.com/Azure/kubelogin#device-code-flow-interactive) I have updated `kubeconfig`:

```yaml
user:
    exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      args:
      - get-token
      - --environment
      - AzurePublicCloud
      - --server-id
      - xxx
      - --client-id
      - xxx
      - --tenant-id
      - xxx
      - --legacy
      command: kubelogin
      env: null
```

After running a kubectl command and confirming my identity I have credentials stored in `~/.kube/cache/kubelogin/azure.json`.
When I tried to run another kubectl command and it asked me to confirm my identity again.

I tried to not use `--legacy` and authenticate again and I got from k8s api server (obviously non-legacy type of credentials is not accepted):
```
error: You must be logged in to the server (the server has asked for the client to provide credentials)
```
When I tried to switch back to `--legacy` and still got the same error message.
The error disappeared when I cleared the cache and reauthenticated but I was back to having to authenticate every `kubectl` execution.

weinong:
thanks for reporting!

PR #6, opened 2020-04-01
fix: cached legacy token is verified correctly

danmx:
Addressing https://github.com/Azure/kubelogin/issues/5

### DIFF
--- a/pkg/token/execCredentialPlugin.go
+++ b/pkg/token/execCredentialPlugin.go
@@ -47,7 +47,11 @@ func (p *execCredentialPlugin) Do() error {
 	}
 
 	// verify resource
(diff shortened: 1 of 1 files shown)


pkg/token/execCredentialPlugin.go
```diff
@@ -47,7 +47,7 @@ func (p *execCredentialPlugin) Do() error {
 	}
 
 	// verify resource
-	if token.Resource == p.o.ServerID && !token.IsZero() {
+	if (token.Resource == p.o.ServerID || token.Resource == fmt.Sprintf("%s:%s", servicePrincipalLogin, p.o.ServerID)) && !token.IsZero() {
```
weinong, line 50:
```suggestion
	targetAudience := p.o.ServerID
	if p.o.IsLegacy {
		targetAudience = fmt.Sprintf("spn:%s", p.o.ServerID)
	}
	if token.Resource == targetAudience && !token.IsZero() {
```
since we know whether it's legacy or not, lets do a proper validation based on legacy flag

Issue #7, opened 2020-04-02
add unit tests


weinong:
let's improve code coverage

PR #10, opened 2020-04-03
added unit tests to convert cmd

weinong:


weinong:
it adds unit tests to converter cmd. #7

PR #16, opened 2020-06-02
Add GetToken as a public method

bingosummer:
Add GetToken as a public method so that "github.com/Azure/kubelogin/pkg/token" can be imported to get token.


bingosummer:
@weinong could you please have a review?

weinong:
if we want to use it as a library, need to ensure it handles concurrency

PR #27, opened 2020-08-14
allow msi login with identityResourceID

haitch:
allow msi login method to use **--identity-resource-id** in addition to current **--client-id**

here is a use case I've tested

```bash
kubelogin get-token --server-id {xxxx} --login msi --identity-resource-id /subscriptions/{subId}/resourceGroups/haitch/providers/Microsoft.ManagedIdentity/userAssignedIdentities/haitch-msi
```


pkg/token/options.go
```diff
@@ -60,6 +61,7 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&o.ClientSecret, "client-secret", o.ClientSecret, fmt.Sprintf("AAD client application secret. Used in spn login. It may be specified in %s environment variable", envServicePrincipalClientSecret))
 	fs.StringVar(&o.Username, "username", o.Username, fmt.Sprintf("user name for ropc login flow. It may be specified in %s environment variable", envROPCUsername))
 	fs.StringVar(&o.Password, "password", o.Password, fmt.Sprintf("password for ropc login flow. It may be specified in %s environment variable", envROPCPassword))
+	fs.StringVar(&o.IdentityResourceId, "identity-resource-id", o.IdentityResourceId, "Managed Identity resource id.")
```
weinong, line 64: can you also update https://github.com/Azure/kubelogin/blob/a7aa1ad9351bd453ace5a2c683c876409c2a63f1/pkg/token/options.go#L103

PR #31, opened 2020-08-20
Add separate zip files per platform

### LINKED ISSUE #25: Separate downloads per platform
Related to #24, I use Windows primarily and to solve https://github.com/lensapp/lens/issues/649 kubelogin was the perfect solution. I looked at the releases tab and aside from the source, a single zip is present. I almost glossed over this as a solution because there is no indication that the build artifact contains Windows, Linux and Mac binaries. It would be helpful to indicate it is for all platforms, or separate these into platform specific files.

Thanks!

ocdi:
Fixes #25 

I tested this on my fork and the releases can be seen here https://github.com/ocdi/kubelogin/releases

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,10 +42,25 @@ jobs:
       uses: montudor/action-zip@v0.1.0
       with:
         args: zip -qq -r kubelogin.zip bin
+
+    - name: Zip (win-amd64)
+      uses: montudor/action-zip@v0.1.0
+      with:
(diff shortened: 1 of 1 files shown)


weinong:
@ocdi can we keep the old behavior (kubelogin.zip) as azure-cli is taking a dependency on it.

ocdi:
ahhh yes absolutely, I wasn't aware of that reason or usage, will update

ocdi:
I restored the original behaviour and this is how the releases look to me now on my test

image

weinong:
that's great!

Issue #41, opened 2020-11-03
"parameter 'identityResourceID' cannot be empty" when using default MSI


adrianwyatt:
Logging in with type "MSI" (i.e. `-l msi`) later fails with `parameter 'identityResourceID' cannot be empty`. These commands are being run on an Azure VMSS with a system-assigned identity.

Repro commands:
1. `az login --identity`
2. `az account set -s {subscription}`
3. `sudo az aks get-credentials --resource-group {resourceGroup} --name {clusterName} --overwrite-existing`
4. `sudo kubelogin convert-kubeconfig -l msi`
5. `sudo kubectl get nodes`

At this point we are seeing this error consistently when using any kubectl commands that access the cluster.
```
Error: failed to get token: failed to create service principal from managed identity  for token refresh: parameter 'identityResourceID' cannot be empty
```

adrianwyatt:
Perhaps the check of `p.identityResourceID == ""` [here ](https://github.com/Azure/kubelogin/blob/master/pkg/token/msi.go#L47) is improper? That would make it fall through to `NewServicePrincipalTokenFromMSIWithIdentityResourceID`

```
if p.clientID == "" {
		if p.identityResourceID == "" {
			// no identity specified, use whatever IMDS default to
			spt, err = adal.NewServicePrincipalTokenFromMSI(
				msiEndpoint,
				p.resourceID,
				callback)
			if err != nil {
				return emptyToken, fmt.Errorf("failed to create service principal from managed identity for token refresh: %s", err)
			}
		}
```

haitch:
can you paste the kubeconfig? you can trim off the sensitive part

haitch:
@adrianwyatt @weinong  can we close this issue?

adrianwyatt:
Looks good to me - thank you.

Sent from my phone.
________________________________

PR #42, opened 2020-11-03
fix a incorrect fallthrough when use msi login

haitch:
after spt created from ```NewServicePrincipalTokenFromMSI```, it invokes ```NewServicePrincipalTokenFromMSIWithIdentityResourceID``` right after without any condition guard, which is wrong, and should be guard with else condition.

### DIFF
--- a/pkg/token/msi.go
+++ b/pkg/token/msi.go
@@ -53,19 +53,18 @@ func (p *managedIdentityToken) Token() (adal.Token, error) {
 			if err != nil {
 				return emptyToken, fmt.Errorf("failed to create service principal from managed identity for token refresh: %s", err)
 			}
-		}
-
-		// use a specified managedIdentity resource id
-		spt, err = adal.NewServicePrincipalTokenFromMSIWithIdentityResourceID(
-			msiEndpoint,
-			p.resourceID,
-			p.identityResourceID,
-			callback)
-		if err != nil {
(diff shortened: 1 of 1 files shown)


haitch:
fixing #41

haitch:
@adrianwyatt I found the issue, the if condition is proper. 

but after we ```NewServicePrincipalTokenFromMSI```, we throw that SPT away and do a ```NewServicePrincipalTokenFromMSIWithIdentityResourceID``` instead, which shouldn't happen.

add an else to fix that.

pkg/token/msi.go
```diff
@@ -53,19 +53,18 @@ func (p *managedIdentityToken) Token() (adal.Token, error) {
...
-			return emptyToken, fmt.Errorf("failed to create service principal from managed identity %s for token refresh: %s", p.identityResourceID, err)
+		} else {
+			// use a specified managedIdentity resource id
+			spt, err = adal.NewServicePrincipalTokenFromMSIWithIdentityResourceID(
+				msiEndpoint,
+				p.resourceID,
+				p.identityResourceID,
+				callback)
+			if err != nil {
+				return emptyToken, fmt.Errorf("failed to create service principal from managed identity %s for token refresh: %s", p.identityResourceID, err)
```
weinong, line 64:
```suggestion
				return emptyToken, fmt.Errorf("failed to create service principal from managed identity with resourceID %s for token refresh: %s", p.identityResourceID, err)
```

weinong (approved):
LGTM

adrianwyatt:
Thank you for the prompt fix!

Issue #52, opened 2021-02-10
Kubectl authorization failing with timeout error in the Docker based Azure App Service


skythet:
I'm using Docker based Azure App Service as a Azure Devops deployment agent for reaching our private network. In this app service enabled [Identity](https://docs.microsoft.com/en-us/azure/app-service/overview-managed-identity?tabs=dotnet#add-a-system-assigned-identity) and I want to use this identity for accessing AKS. 

In AKS granted ClusterAdmin role for this identity:

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: <binding name>
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: cluster-admin
subjects:
- apiGroup: rbac.authorization.k8s.io
  kind: User
  name: <managed identity object id>
```

After configuration I tried to run deployment, but got next error:

```
W0209 12:49:50.510666    1246 transport.go:260] Unable to cancel request for *exec.roundTripper Error: failed to get token: adal: Failed to execute the refresh request. Error = 'Get http://169.254.169.254/metadata/identity/oauth2/token?api-version=2017-09-01&resource=<id>: dial tcp 169.254.169.254:80: i/o timeout' Unable to connect to the server: getting credentials: exec: executable kubelogin failed with exit code 1 (Client.Timeout exceeded while awaiting headers)
```

Kubelogin tries to get token from 169.254.169.254, but this endpoint works only under VM's, for App Services there is available "MSI_ENDPOINT" environment variable and kubelogin should use that endpoint. Also in the kubelogin this endpoint is hardcoded: https://github.com/Azure/kubelogin/blob/master/pkg/token/msi.go#L39

Continuing investigation I've found that  in a newer version of library adal `GetMSIVMEndpoint()` is deprecated and `NewServicePrincipalTokenFromMSI()` can automatically detect correct MSI endpoint: https://github.com/Azure/go-autorest/blob/master/autorest/adal/token.go#L714

skythet:
Pull request merged to master branch

PR #53, opened 2021-02-10
Replacing GetMSIVMEndpoint to NewServicePrincipalTokenFromMSI for avoiding custom msi endpoint

skythet:
Fix for: https://github.com/Azure/kubelogin/issues/52

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,12 +3,12 @@ module github.com/Azure/kubelogin
 go 1.13
 
 require (
-	github.com/Azure/go-autorest/autorest v0.9.0
-	github.com/Azure/go-autorest/autorest/adal v0.9.1
+	github.com/Azure/go-autorest/autorest v0.11.17
+	github.com/Azure/go-autorest/autorest/adal v0.9.12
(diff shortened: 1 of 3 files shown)


weinong:
https://github.com/Azure/go-autorest/blob/79a63e70c4d8b2840383d2a8791e41b6830ca9ad/autorest/adal/token.go#L736

`GetMSIEndpoint()` seems to be deprecated. I believe all we need is bump the adal version and `NewServicePrincipalTokenFromMSI()` will automatically pick up the right MSI endpoint. Can you give that a try?

weinong:
in fact, `NewServicePrincipalTokenFromMSI ()` is also deprecated, would you mind refactoring to use `NewServicePrincipalTokenFromManagedIdentity()`?

skythet:
Thanks, @weinong! I missed that. I've refactored my changes and replaced all the calls to NewServicePrincipalTokenFromManagedIdentity. Also verified changed kubelogin in our infrastructure under app service and in VM, everything works well for us.

weinong (approved):
LGTM. Thanks!

PR #54, opened 2021-02-11
Add Azure CLI as an authentication token source

phealy:
Most people using kubelogin are probably already logged into Azure CLI,
because they probably used az aks get-credentials to build their
kubeconfig in the first place. This login method makes use of the go SDK
for `az account get-access-token` to generate a token whenever required
using the currently logged in account. This means that users only have
to maintain their current login in a single location, instead of two (az
and kubelogin).


weinong:
very nice addition! I will take a look next week!

README.md
```diff
@@ -4,13 +4,14 @@ This is a [client-go credential (exec) plugin](https://kubernetes.io/docs/refere
...
-* non-interactive managed service identity login
-* AAD token will be cached locally for renewal in device code login and user principal login (ropc) flow. By default, it is saved in `~/.kube/cache/kubelogin/`
-* addresses https://github.com/kubernetes/kubernetes/issues/86410 to remove `spn:` prefix in `audience` claim, if necessary. (based on kubeconfig or commandline argument `--legacy`)
+- `convert-kubeconfig` command to converts kubeconfig with existing azure auth provider format to exec credential plugin format
+- [device code login](<#device-code-flow-(interactive)>)
+- [non-interactive service principal login](<#service-principal-login-flow-(non-interactive)>)
+- [non-interactive user principal login](<#user-principal-login-flow-(non-interactive)>) using [Resource owner login flow](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth-ropc)
+- [non-interactive managed service identity login](<#managed-service-identity-(non-interactive)>)
+- [non-interactive Azure CLI token login](<#azure-cli-token-login-(non-interactive)>)
```
weinong, line 12: since we are adopting kubelogin for non managed Kubernetes on Azure as well, I think we should be clear that azure cli login will only work for AKS unless additional configuration is done?

README.md
```diff
@@ -156,6 +155,18 @@ kubelogin convert-kubeconfig -l msi --client-id msi-client-id
 kubectl get no
 ```
 
+#### Azure CLI token login (non interactive)
```
weinong, line 158: suggestion: add a note saying it's for AKS only or provide instruction how to setup for customer's own AAD app.

README.md
```diff
@@ -324,26 +335,52 @@ users:
...
+      exec:
+        apiVersion: client.authentication.k8s.io/v1beta1
+        args:
+          - get-token
+          - --environment
+          - AzurePublicCloud
+          - --server-id
+          - <Server_Appid>
+          - --client-id
+          - <Client_Appid>
```
weinong, line 370: I believe you won't need client app ID

README.md
phealy: If you can tell me what other configuration would be needed, I can easily add that in. I didn't realize that it wouldn't work for non-managed K8s.

README.md
phealy: As I replied above - happy to add this in, I just didn't know it would be required. Do you have a pointer to doc on what would need to be added? If not, I'll try to figure it out Monday.

pkg/token/options.go
```diff
@@ -31,6 +31,7 @@ const (
 	ROPCLogin             = "ropc"
 	MSILogin              = "msi"
 	manualTokenLogin      = "manual_token"
+	azureCLILogin         = "azurecli"
```
weinong, line 34:
```suggestion
	AzureCLILogin         = "azurecli"
```

weinong: this is to match other login methods. in fact, let's insert it at L33?

pkg/token/azurecli.go
```diff
@@ -0,0 +1,72 @@
...
+	"fmt"
+	"strconv"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/Azure/go-autorest/autorest/adal"
+)
+
+type azureCLIToken struct {
+	clientID    string
```
weinong, line 16: clientID is not used. please remove it

weinong: what happens is, it's using the azure-cli's client instead.

weinong: since clientID is not used, can you also remove it in `convert.go`?

README.md
weinong: I just played around with that, and I don't think it's possible. so AKS only :)

weinong:
what is the experience like when azure cli is not installed?

README.md
phealy: Note added to the description.

pkg/token/azurecli.go
phealy: Actually, tenant ID wasn't used either, because you can only be signed in using one tenant. Removed that too, all the way through.

README.md
```diff
@@ -156,6 +155,18 @@ kubelogin convert-kubeconfig -l msi --client-id msi-client-id
...
+
+```sh
+export KUBECONFIG=/path/to/kubeconfig
+
+kubelogin convert-kubeconfig -l azurecli
+
+kubectl get no
+```
+
+Uses an [access token](https://docs.microsoft.com/en-us/cli/azure/account?view=azure-cli-latest#az_account_get_access_token) from [Azure CLI](https://github.com/Azure/azure-cli) to log in. The token will be issued against whatever tenant was logged in at the time `kubelogin convert-kubeconfig -l azurecli` was run. This login option only works with managed AKS.
```
weinong, line 168:
```suggestion
Uses an [access token](https://docs.microsoft.com/en-us/cli/azure/account?view=azure-cli-latest#az_account_get_access_token) from [Azure CLI](https://github.com/Azure/azure-cli) to log in. The token will be issued against whatever tenant was logged in at the time `kubelogin convert-kubeconfig -l azurecli` was run. This login option only works with managed AAD in AKS.

```

weinong (approved):
LGTM

PR #63, opened 2021-06-08
BUGFIX: Refresh token is removed from credentials when it is used

maxburke:
The observed behavior is that when the access token expires and the refresh token is used to fetch a new access token, the refresh token is lost. The result is that when the access token expires again, there is no refresh token and the user is prompted to re-authenticate. This becomes a huge pain point when people are trying to do a days work because it means having to re-authenticate several times per day.

This change adds the refresh token back to the token object and allows the user to successively refresh their access token until the refresh token expires.

### DIFF
--- a/pkg/token/execCredentialPlugin.go
+++ b/pkg/token/execCredentialPlugin.go
@@ -85,12 +85,18 @@ func (p *execCredentialPlugin) Do() error {
 				return fmt.Errorf("failed to get refresher: %s", err)
 			}
 			klog.V(5).Info("refresh token")
+			refreshToken := token.RefreshToken
 			token, err := refresher.Token()
+
 			// if refresh fails, we will login using token provider
 			if err != nil {
 				klog.V(5).Infof("refresh failed, will continue to login: %s", err)
 			} else {
 				tokenRefreshed = true
+			}
+
+			if token.RefreshToken == "" {
+				token.RefreshToken = refreshToken
 			}
 
 			if tokenRefreshed {


pkg/token/execCredentialPlugin.go
```diff
@@ -85,14 +85,20 @@ func (p *execCredentialPlugin) Do() error {
...
 			token, err := refresher.Token()
+
 			// if refresh fails, we will login using token provider
 			if err != nil {
 				klog.V(5).Infof("refresh failed, will continue to login: %s", err)
 			} else {
 				tokenRefreshed = true
 			}
 
+			if token.RefreshToken == "" {
```
weinong, line 98: do you happen to know under what condition `token.RefreshToken` would be empty? what was the auth flow being used?

maxburke: Hi @weinong !

I don't know the specific auth flow (if you're looking for the flow ID). We have our Kubernetes cluster setup to authenticate against Azure AD. One possible sticking point is that our Azure AD is setup to delegate to Google for SSO.

One thing we have been noticing for most users (created after January 2020, mostly) is that whenever their access token expires and is refreshed, the token that comes back has no refresh token field which makes it really difficult for people to do their work, as once that refreshed access token expires they need to do the device code sign-in again.

weinong: @maxburke would it be possible to send me the fiddler trace showing the empty refresh token? I can engage with AAD team to take a look. you can send that to me at weinongw at microsoft.com :)

weinong: @maxburke if you cannot send me the repro, can you at least add a log line such as
```
klog.V(10).Info("after token refresh refresh_token is still empty, reuse the existing refresh_token...")
```

weinong (changes requested):
please add the requested log line

PR #82, opened 2022-03-29
fix: store resource id to allow reusing cached token for azure cli

lnr0626:
I was investigating some odd performance issues when using this and found out that the cached token value was never used because its resource id was always an empty string and so it would never match the requested resource id.


weinong (approved):
thank you for the contribution!

PR #86, opened 2022-04-09
Added retry around writing cache

### LINKED ISSUE #71: Access is denied to move temporary token to desired output location
I am using Kubelogin plugin and run command `kubelogin convert-kubeconfig -l azurecli` to refresh aks token. but got the following error:

image

Any direction on how I can grant the permission or any insight on the issue?

It does successfully refresh the aks token but sometimes the kubectl commands are not working

ITHedgeHog:
Fixes #71 by adding a retry around writing to the token cache. It retries the write every 250 milliseconds for 1 second.


ITHedgeHog:
@weinong We've been running this for the last week within the team and have not had this error happen once.

Issue #93, opened 2022-07-01
Kubelogin is not able to convert the kubeconfig to azurecli format for 1.24.0 AKS cluster


gyuvaraj10:
Step Performed:
1. az login
2. az aks get-credentials -g "abc" -n  "def"
3. kubelogin convert-kubeconfig -l azurecli

Actual config in the .kube/config file
```
- name: clusterUser_rc-a-aks-1
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      args:
      - get-token
      - --environment
      - AzurePublicCloud
      - --server-id
      - xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
      - --client-id
      - xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
      - --tenant-id
      - xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
      - --login
      - devicecode
      command: kubelogin
      env: null
      provideClusterInfo: false
```

Expected Config:
```
- name: clusterUser_rc-zaks-1
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      args:
      - get-token
      - --server-id
      - xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
      - --login
      - azurecli
      command: kubelogin
      env: null
      provideClusterInfo: false
```

**Environment (please complete the following information):**
 -  Kubectl Client CLI Version - 1.24.1
 - Kubernetes version [e.g. 1.24.0]
 -  Kubelogin - 
  git hash: v0.0.13/52e83a071f39e9e039e95aa9a6fbea04855eae13
 Go version: go1.17.9
 Build time: 2022-04-23T00:49:14Z

slnowak:
+1, discovered same issue today after upgrading to 1.24.0

weinong:
since 1.24.0, AKS default returns exec format kubeconfig so that it doesn't actually need kubelogin conversion. currently, kubelogin's `convert-kubeconfig` sub command doesn't do anything unless the kubeconfig is pre 1.24 format using azure auth. one can argue that `convert-kubeconfig` should probably convert the exec format as well to support different login method.

weinong:
@gyuvaraj10, the kubeconfig you got from 1.24 cluster should be as good unless you require to use azurecli login?

gyuvaraj10:
[...]

PR #94, opened 2022-07-06
AKS 1.24.0: kubeconfig to azurecli

anmathew:
Addresses #93

### DIFF
--- a/pkg/converter/convert.go
+++ b/pkg/converter/convert.go
@@ -130,9 +130,6 @@ func Convert(o Options) error {
 				exec.Args = append(exec.Args, serveridArg)
 				exec.Args = append(exec.Args, argLoginMethod)
 				exec.Args = append(exec.Args, o.TokenOptions.LoginMethod)
-				authInfo.Exec = exec
-				authInfo.AuthProvider = nil
-				continue
 			} else {
 				// others are not supported yet
 				fmt.Fprintln(os.Stderr, "err: not supported yet")
@@ -194,9 +191,9 @@ func Convert(o Options) error {
 				exec.Args = append(exec.Args, argLoginMethod)
 				exec.Args = append(exec.Args, o.TokenOptions.LoginMethod)
 			}
-			authInfo.Exec = exec
-			authInfo.AuthProvider = nil
 		}
+		authInfo.Exec = exec
+		authInfo.AuthProvider = nil
 	}
 
 	clientcmd.ModifyConfig(clientcmd.NewDefaultPathOptions(), config, true)
--- a/pkg/converter/convert.go
+++ b/pkg/converter/convert.go
@@ -56,73 +56,91 @@ func Convert(o Options) error {
 	isAlternativeLogin := isMSI || isAzureCLI || isWorkloadIdentity
 
 	for _, authInfo := range config.AuthInfos {
+		exec := &api.ExecConfig{
+			Command: execName,
+			Args: []string{
+				getTokenCommand,
+			},
+			APIVersion: execAPIVersion,
+		}
+
 		if authInfo != nil {
+
 			if authInfo.AuthProvider == nil || authInfo.AuthProvider.Name != azureAuthProvider {
-				continue
+				if isAzureCLI && !o.TokenOptions.IsLegacy {
+					exec.Args = append(exec.Args, argServerID)
+					serveridArg := o.TokenOptions.ServerID
+					if len(serveridArg) < 1 {
+						serveridArg = "6dae42f8-4368-4678-94ff-3960e28e3630"
+					}
+					exec.Args = append(exec.Args, serveridArg)
+					exec.Args = append(exec.Args, argLoginMethod)
+					exec.Args = append(exec.Args, o.TokenOptions.LoginMethod)
+					authInfo.Exec = exec
+					authInfo.AuthProvider = nil
+					break
+				} else {
+					continue
+				}
+			} else {
+
+				if !isAlternativeLogin && o.isSet(flagEnvironment) {
+					exec.Args = append(exec.Args, argEnvironment)
+					exec.Args = append(exec.Args, o.TokenOptions.Environment)
+				} else if !isAlternativeLogin && authInfo.AuthProvider.Config[cfgEnvironment] != "" {
+					exec.Args = append(exec.Args, argEnvironment)
+					exec.Args = append(exec.Args, authInfo.AuthProvider.Config[cfgEnvironment])
+				}
+				if o.isSet(flagServerID) {
+					exec.Args = append(exec.Args, argServerID)
+					exec.Args = append(exec.Args, o.TokenOptions.ServerID)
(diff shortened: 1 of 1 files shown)


pkg/converter/convert.go
```diff
@@ -56,73 +56,91 @@ func Convert(o Options) error {
...
-				exec.Args = append(exec.Args, o.TokenOptions.Username)
-			}
-			if !isAlternativeLogin && o.isSet(flagPassword) {
-				exec.Args = append(exec.Args, argPassword)
-				exec.Args = append(exec.Args, o.TokenOptions.Password)
-			}
-			if o.isSet(flagLoginMethod) {
-				exec.Args = append(exec.Args, argLoginMethod)
-				exec.Args = append(exec.Args, o.TokenOptions.LoginMethod)
+				if isAzureCLI && !o.TokenOptions.IsLegacy {
```
weinong, line 70: we should check if this is exec plugin using kubelogin instead of blindly convert

pkg/converter/convert.go
```diff
@@ -56,73 +56,91 @@ func Convert(o Options) error {
...
+					exec.Args = append(exec.Args, argServerID)
+					serveridArg := o.TokenOptions.ServerID
+					if len(serveridArg) < 1 {
+						serveridArg = "6dae42f8-4368-4678-94ff-3960e28e3630"
```
weinong, line 74: serverID should come from the argument given to existing kubelogin command

pkg/converter/convert.go
```diff
@@ -54,75 +78,93 @@ func Convert(o Options) error {
...
 	isWorkloadIdentity := o.TokenOptions.LoginMethod == token.WorkloadIdentityLogin
 	isAlternativeLogin := isMSI || isAzureCLI || isWorkloadIdentity
-
 	for _, authInfo := range config.AuthInfos {
+
 		if authInfo != nil {
-			if authInfo.AuthProvider == nil || authInfo.AuthProvider.Name != azureAuthProvider {
-				continue
-			}
+
```
weinong, line 84: i think this is where we should check: is it legacy aad auth or is it exec using kubelogin? if neither, we return and do nothing

pkg/converter/convert.go
```diff
@@ -54,75 +78,93 @@ func Convert(o Options) error {
...
+					serveridArg := getServerId(authInfo)
+					if len(serveridArg) < 1 {
+						serveridArg = "6dae42f8-4368-4678-94ff-3960e28e3630"
+					}
+					exec.Args = append(exec.Args, serveridArg)
+					exec.Args = append(exec.Args, argLoginMethod)
+					exec.Args = append(exec.Args, o.TokenOptions.LoginMethod)
+					authInfo.Exec = exec
+					authInfo.AuthProvider = nil
+					break
```
weinong, line 105: the existing code converts all of the authInfo. Let's keep it that way

pkg/converter/convert.go
```diff
@@ -43,6 +43,30 @@ const (
...
+		}
+	}
+	return
+}
+
+func isUsingKubelogin(execConfigPtr *api.ExecConfig) (ok bool) {
+	if execConfigPtr == nil {
+		return
+	}
+	return execConfigPtr.Command == "kubelogin"
```
weinong, line 67: i think we should check if Command contains kubelogin as it may be ./kubelogin, /path/to/kubelogin, or even kubelogin.exe

pkg/converter/convert.go
```diff
@@ -43,6 +43,30 @@ const (
...
 
+func getServerId(authInfoPtr *api.AuthInfo) (serverId string) {
+	if authInfoPtr == nil || authInfoPtr.Exec == nil || authInfoPtr.Exec.Args == nil {
+		return
+	}
+	if len(authInfoPtr.Exec.Args) < 1 {
+		return
+	}
+	for i := range authInfoPtr.Exec.Args {
+		if authInfoPtr.Exec.Args[i] == "--server-id" {
```
weinong, line 54:
```suggestion
		if authInfoPtr.Exec.Args[i] == argServerID {
```

pkg/converter/convert.go
```diff
@@ -80,91 +104,98 @@ func Convert(o Options) error {
 	isAlternativeLogin := isMSI || isAzureCLI || isWorkloadIdentity
 	for _, authInfo := range config.AuthInfos {
 
-		if authInfo != nil {
+		//  is it legacy aad auth or is it exec using kubelogin?
+		if isExecUsingkubelogin(authInfo) || isLegacyAADAuth(authInfo) {
```
weinong, line 108:
```suggestion
		if !isExecUsingkubelogin(authInfo) && !isLegacyAADAuth(authInfo) {
		    continue
		}
```

pkg/converter/convert.go
```diff
@@ -80,91 +104,98 @@ func Convert(o Options) error {
...
-					exec.Args = append(exec.Args, o.TokenOptions.Password)
-				}
-				if o.isSet(flagLoginMethod) {
-					exec.Args = append(exec.Args, argLoginMethod)
-					exec.Args = append(exec.Args, o.TokenOptions.LoginMethod)
+			if isAzureCLI { //support azurecli login for now
+				exec.Args = append(exec.Args, argServerID)
+				serveridArg := getServerId(authInfo)
+				if len(serveridArg) < 1 {
+					fmt.Fprintln(os.Stderr, "err: serverId")
```
weinong, line 127: return error and let the caller log the error and exit

pkg/converter/convert.go
```diff
@@ -80,91 +104,98 @@ func Convert(o Options) error {
...
-					exec.Args = append(exec.Args, argPassword)
-					exec.Args = append(exec.Args, o.TokenOptions.Password)
-				}
-				if o.isSet(flagLoginMethod) {
-					exec.Args = append(exec.Args, argLoginMethod)
-					exec.Args = append(exec.Args, o.TokenOptions.LoginMethod)
+			if isAzureCLI { //support azurecli login for now
+				exec.Args = append(exec.Args, argServerID)
+				serveridArg := getServerId(authInfo)
+				if len(serveridArg) < 1 {
```
weinong, line 126: > if len(serveridArg) < 1

it looks kinda weird. either checking against empty string or 0 length..

pkg/converter/convert.go
```diff
@@ -80,91 +104,98 @@ func Convert(o Options) error {
...
 				}
+				exec.Args = append(exec.Args, serveridArg)
+				exec.Args = append(exec.Args, argLoginMethod)
+				exec.Args = append(exec.Args, o.TokenOptions.LoginMethod)
 				authInfo.Exec = exec
 				authInfo.AuthProvider = nil
+				continue
+			} else {
+				// others are not supported yet
+				fmt.Fprintln(os.Stderr, "err: not supported yet")
```
weinong, line 138: same here: return fmt.Errorf("%q is not supported yet", o.TokenOptions.LoginMethod)

weinong (approved):
LGTM

PR #122, opened 2022-08-06
pass in tenantID to azurecli auth

weinong:
pass in tenantID to azurecli auth


pearj:
I tested this locally and it works great. I can now sign into multiple azure tenancies and azure cli auth continues to work when I switch between AKS clusters in different tenancies.

pearj:
It turns out it's a little more complicated. When a token expires it can't fetch a new one unless the current tenant is set with `az account set --subscription`. I guess I'll need to have a bash script that switches the azure cli primary subscription when I change the kubernetes context.
But I think it's still workable. I don't think it's something that `kubelogin` can fix.

PR #126, opened 2022-08-22
Support encrypted client certificates for service principal login

sybereal:
The decryption password for reading the PKCS#12 file was previously hardcoded as the empty string. This made it impossible to use client certificates that have a password set, which was especially inconvenient in the context of Terraform, as the Azure Terraform providers do support this.

### DIFF
--- a/pkg/converter/convert.go
+++ b/pkg/converter/convert.go
@@ -23,6 +23,7 @@ const (
 	argEnvironment        = "--environment"
 	argClientSecret       = "--client-secret"
 	argClientCert         = "--client-certificate"
+	argClientCertPassword = "--client-certificate-password"
 	argIsLegacy           = "--legacy"
 	argUsername           = "--username"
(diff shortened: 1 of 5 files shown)


pkg/converter/convert.go
```diff
@@ -247,6 +249,11 @@ func Convert(o Options, pathOptions *clientcmd.PathOptions) error {
 				exec.Args = append(exec.Args, o.TokenOptions.ClientCert)
 			}
 
+			if o.isSet(flagClientCertPassword) {
```
weinong, line 252: can you also add a test case in convert_test.go

weinong:
Hi @sybereal, LGTM. Thanks for your contribution! Please add a test case and we can merge it!

sybereal:
Thank you for the feedback, @weinong. I have added the test case as requested.

weinong (approved):
thank you!

PR #132, opened 2022-09-13
Add how to write secret permantly into kubeconfig

Xaseron:
In Azure DevOps setting environment variables for multiple jobs is PITA.  Writing the client id and secret into the kubeconfig makes it easier.


weinong (approved):
thanks for your contribution!

PR #133, opened 2022-09-14
Update objectId to id on Line 90

everjing:
`az ad sp show --id <spn client id> --query "objectId"` is wrong. Should be `az ad sp show --id <spn client id> --query "id"' for service principal AAD Object ID


weinong (approved):
thanks!

PR #135, opened 2022-09-25
Add webbrowser login method

everjing:
Added webbrowser login method for kubelogin. kubeconfig works and web browser opens when calling `kubectl get no` but anthentication failed with error code AADSTS65002.

### DIFF
--- a/pkg/converter/convert.go
+++ b/pkg/converter/convert.go
@@ -217,6 +217,28 @@ func Convert(o Options, pathOptions *clientcmd.PathOptions) error {
 				exec.Args = append(exec.Args, argIsLegacy)
 			}
 
+		case token.InteractiveLogin:
+
+			if argClientIDVal == "" {
+				return fmt.Errorf("%s is required", argClientID)
+			}
+
+			exec.Args = append(exec.Args, argClientID)
+			exec.Args = append(exec.Args, argClientIDVal)
+
+			if argTenantIDVal == "" {
+				return fmt.Errorf("%s is required", argTenantID)
+			}
+
+			exec.Args = append(exec.Args, argTenantID)
+			exec.Args = append(exec.Args, argTenantIDVal)
+
+			if argEnvironmentVal != "" {
+				// environment is optional
+				exec.Args = append(exec.Args, argEnvironment)
+				exec.Args = append(exec.Args, argEnvironmentVal)
+			}
+
 		case token.ServicePrincipalLogin:
 
 			if argClientIDVal == "" {
--- a/pkg/token/interactive.go
+++ b/pkg/token/interactive.go
@@ -1,0 +1,83 @@
+package token
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strconv"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/cloud"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/Azure/go-autorest/autorest/adal"
+)
+
+type InteractiveToken struct {
+	clientID    string
+	resourceID  string
+	tenantID    string
+	oAuthConfig adal.OAuthConfig
+}
+
+// newInteractiveTokenProvider returns a TokenProvider that will fetch a token for the user currently logged into the Interactive.
+// Required arguments include an oAuthConfiguration object and the resourceID (which is used as the scope)
+func newInteractiveTokenProvider(oAuthConfig adal.OAuthConfig, clientID, resourceID, tenantID string) (TokenProvider, error) {
+	if clientID == "" {
+		return nil, errors.New("clientID cannot be empty")
+	}
+	if resourceID == "" {
+		return nil, errors.New("resourceID cannot be empty")
+	}
+	if tenantID == "" {
+		return nil, errors.New("tenantID cannot be empty")
+	}
+
+	return &InteractiveToken{
+		clientID:    clientID,
+		resourceID:  resourceID,
+		tenantID:    tenantID,
+		oAuthConfig: oAuthConfig,
+	}, nil
+}
+
+// Token fetches an azcore.AccessToken from the interactive browser SDK and converts it to an adal.Token for use with kubelogin.
+func (p *InteractiveToken) Token() (adal.Token, error) {
+	emptyToken := adal.Token{}
+
+	// Request a new Interactive token provider
+	authorityFromConfig := p.oAuthConfig.AuthorityEndpoint
+	clientOpts := azcore.ClientOptions{Cloud: cloud.Configuration{
+		ActiveDirectoryAuthorityHost: authorityFromConfig.String(),
+	}}
+	cred, err := azidentity.NewInteractiveBrowserCredential(&azidentity.InteractiveBrowserCredentialOptions{
+		ClientOptions: clientOpts,
+		TenantID:      p.tenantID,
+		ClientID:      p.clientID,
+	})
+	if err != nil {
+		return emptyToken, fmt.Errorf("unable to create credential. Received: %w", err)
+	}
+
+	// Use the token provider to get a new token
+	interactiveToken, err := cred.GetToken(context.Background(), policy.TokenRequestOptions{Scopes: []string{p.resourceID + "/.default"}})
+	if err != nil {
+		return emptyToken, fmt.Errorf("expected an empty error but received: %w", err)
+	}
+	if len(interactiveToken.Token) == 0 {
+		return emptyToken, errors.New("did not receive a token")
+	}
+
+	// azurecore.AccessTokens have ExpiresOn as Time.Time. We need to convert it to JSON.Number
+	// by fetching the time in seconds since the Unix epoch via Unix() and then converting to a
+	// JSON.Number via formatting as a string using a base-10 int64 conversion.
+	expiresOn := json.Number(strconv.FormatInt(interactiveToken.ExpiresOn.Unix(), 10))
+
+	// Re-wrap the azurecore.AccessToken into an adal.Token
+	return adal.Token{
+		AccessToken: interactiveToken.Token,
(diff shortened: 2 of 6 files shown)


pkg/converter/convert.go
```diff
@@ -217,6 +217,32 @@ func Convert(o Options, pathOptions *clientcmd.PathOptions) error {
...
+			exec.Args = append(exec.Args, argTenantID)
+			exec.Args = append(exec.Args, argTenantIDVal)
+
+			if argEnvironmentVal != "" {
+				// environment is optional
+				exec.Args = append(exec.Args, argEnvironment)
+				exec.Args = append(exec.Args, argEnvironmentVal)
+			}
+
+			if isLegacyConfigMode {
```
weinong, line 242: we can get rid of legacy mode

pkg/converter/convert.go
```diff
@@ -217,6 +217,32 @@ func Convert(o Options, pathOptions *clientcmd.PathOptions) error {
 				exec.Args = append(exec.Args, argIsLegacy)
 			}
 
+		case token.WebBrowserLogin:
+
+			if argClientIDVal == "" {
+				return fmt.Errorf("%s is required", argClientID)
+			}
+
+			exec.Args = append(exec.Args, argClientID)
```
weinong, line 226: you are missing server ID

pkg/token/webbrowser.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+type WebBrowserToken struct {
+	clientID    string
+	resourceID  string
+	tenantID    string
+	oAuthConfig adal.OAuthConfig
+}
+
+// newAzureCLIToken returns a TokenProvider that will fetch a token for the user currently logged into the Azure CLI.
```
weinong, line 22: copy paste :)

pkg/converter/convert.go
everjing: The code already added server ID for all the mode as shown in the image below. 
image

pkg/token/webbrowser.go
```diff
@@ -0,0 +1,76 @@
...
+		oAuthConfig: oAuthConfig,
+	}, nil
+}
+
+// Token fetches an azcore.AccessToken from the Azure CLI SDK and converts it to an adal.Token for use with kubelogin.
+func (p *WebBrowserToken) Token() (adal.Token, error) {
+	emptyToken := adal.Token{}
+
+	// Request a new Azure CLI token provider
+	cred, err := azidentity.NewInteractiveBrowserCredential(&azidentity.InteractiveBrowserCredentialOptions{
```
weinong, line 48: you actually need to configure Authority so that it would work in other clouds.
Example: https://github.com/denisenkom/go-mssqldb/blob/master/azuread/configuration.go#L150

weinong: let's configure the authority in newWebBrowerTokenProvider

pkg/token/webbrowser.go
```diff
@@ -0,0 +1,76 @@
...
+		TenantID: p.tenantID,
+		ClientID: p.clientID,
+	})
+	if err != nil {
+		return emptyToken, fmt.Errorf("unable to create credential. Received: %v", err)
+	}
+
+	// Use the token provider to get a new token
+	WebBrowserToken, err := cred.GetToken(context.Background(), policy.TokenRequestOptions{Scopes: []string{p.resourceID}})
```
weinong, line 57: local variable shouldn't start with upper case

pkg/token/options.go
```diff
@@ -60,7 +61,7 @@ var (
 )
 
 func init() {
-	supportedLogin = []string{DeviceCodeLogin, ServicePrincipalLogin, ROPCLogin, MSILogin, AzureCLILogin, WorkloadIdentityLogin}
+	supportedLogin = []string{DeviceCodeLogin, WebBrowserLogin, ServicePrincipalLogin, ROPCLogin, MSILogin, AzureCLILogin, WorkloadIdentityLogin}
```
weinong, line 64: let's call it InteractiveLogin

pkg/token/webbrowser.go
weinong: I found the bug: the resourceID needs to have `/.default` suffix. refer to https://github.com/Azure/azure-sdk-for-go/blob/c5358ab6c0bf824403f0ecdbd164c882d794793d/sdk/azidentity/MIGRATION.md#azidentity-5

pkg/converter/convert.go
```diff
@@ -217,6 +217,32 @@ func Convert(o Options, pathOptions *clientcmd.PathOptions) error {
...
+			if argTenantIDVal == "" {
+				return fmt.Errorf("%s is required", argTenantID)
+			}
+
+			exec.Args = append(exec.Args, argTenantID)
+			exec.Args = append(exec.Args, argTenantIDVal)
+
+			if argEnvironmentVal != "" {
+				// environment is optional
+				exec.Args = append(exec.Args, argEnvironment)
```
bcho, line 238: nit: we can combine the append arguments into one:

```
exec.Args = append(
  exec.Args,
  argEnvironment, argEnvironmentVal,
)
```

pkg/token/webbrowser.go
```diff
@@ -0,0 +1,76 @@
...
+	if err != nil {
+		return emptyToken, fmt.Errorf("expected an empty error but received: %v", err)
```
bcho, line 59: nit: we can use `%w` to wrap the error

ref: https://go.dev/blog/go1.13-errors

pkg/token/webbrowser.go
weinong: we need one additional flags:
* `--cloud` so that you can setup the `azcore.ClientOptions.Cloud`. [ref](https://github.com/Azure/azure-sdk-for-go/blob/c5358ab6c0bf824403f0ecdbd164c882d794793d/sdk/azidentity/MIGRATION.md#azidentity)

besides, we need to take `--authority-host` into account:
* when `--authority-host` is set, we will use it instead of setting `azcore.ClientOptions.Cloud`
* probably should do a validation to check if `--authority-host` and the computed authority host from `--cloud` are the same when both are set

pkg/converter/convert.go
everjing: Done for next commit.

pkg/token/webbrowser.go
everjing: Addressed these two concerns in next commit.

pkg/token/webbrowser.go
everjing: Acked and used w% for next commit.

pkg/token/webbrowser.go
everjing: Updated for the next commit. Thanks!

pkg/token/options.go
everjing: Done for next commit.

pkg/token/webbrowser.go
everjing: Done in the most recent commit f8e875a. Please check it out. Thanks!

pkg/converter/convert.go
```diff
@@ -217,6 +217,28 @@ func Convert(o Options, pathOptions *clientcmd.PathOptions) error {
 				exec.Args = append(exec.Args, argIsLegacy)
 			}
 
+		case token.InteractiveLogin:
```
weinong, line 220: add unit test please

pkg/token/interactive.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// Token fetches an azcore.AccessToken from the interactive browser SDK and converts it to an adal.Token for use with kubelogin.
+func (p *InteractiveToken) Token() (adal.Token, error) {
+	emptyToken := adal.Token{}
+
+	// Request a new Interactive token provider
+	authorityFromConfig := p.oAuthConfig.AuthorityEndpoint
+	clientOpts := azcore.ClientOptions{Cloud: cloud.Configuration{
+		ActiveDirectoryAuthorityHost: authorityFromConfig.String(), Services: map[cloud.ServiceName]cloud.ServiceConfiguration{},
```
weinong, line 52: you don't need to specify `Services`

pkg/token/interactive.go
```diff
@@ -0,0 +1,83 @@
...
+	}}
+	cred, err := azidentity.NewInteractiveBrowserCredential(&azidentity.InteractiveBrowserCredentialOptions{
+		ClientOptions: clientOpts,
+		TenantID:      p.tenantID,
+		ClientID:      p.clientID,
+	})
+	if err != nil {
+		return emptyToken, fmt.Errorf("unable to create credential. Received: %v", err)
```
weinong, line 60:
```suggestion
		return emptyToken, fmt.Errorf("unable to create credential. Received: %w", err)
```

pkg/token/options.go
```diff
@@ -34,6 +34,7 @@ const (
 	defaultEnvironmentName = "AzurePublicCloud"
 
 	DeviceCodeLogin       = "devicecode"
+	InteractiveLogin      = "interactive"
```
weinong, line 37: i'm actualy torn between this or `interactivebrowser`. @bcho WDYT?

bcho: I think we better to follow other SDK / docs to use `interactive`. Example: https://learn.microsoft.com/en-us/python/api/msal/msal.application.publicclientapplication?view=azure-python#msal-application-publicclientapplication-acquire-token-interactive

pkg/converter/convert.go
everjing: Working on it and unit test will show up in next commit.

pkg/token/interactive.go
everjing: OK, will remove Services in next commit.

pkg/token/interactive.go
everjing: Done for next commit. Thanks!

pkg/converter/convert.go
everjing: Added unit test for interactive cases in convert_test.go in the most recent commit e916641. Please check it out. Thanks!

weinong (approved):
LGTM

weinong:
please squash and cleanup the commit history!

PR #138, opened 2022-10-13
Docs: Add appending slash to AAD oidc-issuer-url

epleterte:
When setting this up, I got authenticated (with AAD), but had no further access in Kubernetes. Looking at the logs from kube-apiserver, I could easily pick out error messages complaining about the OIDC issuer URL not being correct. Adding a slash at the end sorted this out.

I hope this saves someone else a few seconds of debugging.


weinong (approved):
LGTM

PR #140, opened 2022-10-17
Add helperGetApiVersionFromEnv() to get apiVersion from KUBERNETES_EX…

everjing:
Got this error message when run 'kubectl get no' with this build locally. 
image


everjing:
With this commit the build goes well with v1beta1, and I was trying to test a scenario using v1 version but it seems even the most recent aks version (1.24.6) is still using v1beta1. So I did not find a way to test v1 cases with the code.

pkg/token/execCredentialWriter.go
```diff
@@ -20,20 +39,67 @@ type execCredentialWriter struct{}
...
+	//os.Setenv(execInfoEnv, "TEST")
+	//fmt.Println("KUBERNETES_EXEC_INFO:", os.Getenv(execInfoEnv))
+	apiVersionFromEnv, err := helperGetApiVersionFromEnv()
+	if err != nil {
+		return err
 	}
+
+	var ec interface{}
+	t := metav1.NewTime(token.Expires())
+	if apiVersionFromEnv == "client.authentication.k8s.io/v1beta1" {
```
weinong, line 52: use a switch to determine the api version. the rest of the code can be outside the switch block to avoid duplication

pkg/token/execCredentialWriter.go
```diff
@@ -20,20 +39,67 @@ type execCredentialWriter struct{}
...
 	if err := e.Encode(ec); err != nil {
 		return fmt.Errorf("could not write the ExecCredential: %s", err)
 	}
 	return nil
 }
+
+func helperGetApiVersionFromEnv() (string, error) {
+	env := os.Getenv(execInfoEnv)
+	if env == "" {
+		return "client.authentication.k8s.io/v1beta1", nil
```
weinong, line 86: use const

pkg/token/execCredentialWriter.go
```diff
@@ -20,20 +39,67 @@ type execCredentialWriter struct{}
...
+		if err := scheme.Convert(obj, &execCredential, nil); err != nil {
+			return "", fmt.Errorf("cannot convert to ExecCredential: %w", err)
+		}
+		if execCredential.TypeMeta.APIVersion == "" {
+			return "client.authentication.k8s.io/v1beta1", nil
+		}
+		if execCredential.TypeMeta.APIVersion == "client.authentication.k8s.io/v1beta1" || execCredential.TypeMeta.APIVersion == "client.authentication.k8s.io/v1" {
+			return execCredential.TypeMeta.APIVersion, nil
+		} else {
+			return "", errors.New("This api version is not supported")
```
weinong, line 102: error message should indicate the provided api version

pkg/token/execCredentialWriter.go
```diff
@@ -20,20 +39,67 @@ type execCredentialWriter struct{}
...
+	} else {
+		obj, _, err := codecs.UniversalDeserializer().Decode([]byte(env), nil, nil)
+		if err != nil {
+			return "", fmt.Errorf("decode: %w", err)
+		}
+		var execCredential clientauthentication.ExecCredential
+		if err := scheme.Convert(obj, &execCredential, nil); err != nil {
+			return "", fmt.Errorf("cannot convert to ExecCredential: %w", err)
+		}
+		if execCredential.TypeMeta.APIVersion == "" {
```
weinong, line 96: switch will look cleaner

pkg/token/execCredentialWriter.go
```diff
@@ -20,20 +28,63 @@ type execCredentialWriter struct{}
...
+	}
+
 	e := json.NewEncoder(os.Stdout)
 	if err := e.Encode(ec); err != nil {
 		return fmt.Errorf("could not write the ExecCredential: %s", err)
 	}
 	return nil
 }
+
+func helperGetApiVersionFromEnv() (string, error) {
```
weinong, line 70: nit: you don't need helper prefix. how about `getAPIVersionFromExecInfoEnv`

pkg/token/execCredentialWriter.go
```diff
@@ -20,20 +28,63 @@ type execCredentialWriter struct{}
...
+	env := os.Getenv(execInfoEnv)
+	if env == "" {
+		return apiV1beta1, nil
+	}
+	var execCredential clientauthentication.ExecCredential
+	error := json.Unmarshal([]byte(env), &execCredential)
+	if error != nil {
+		return "", fmt.Errorf("cannot convert to ExecCredential: %w", error)
```
weinong, line 78:
```suggestion
		return "", fmt.Errorf("cannot unmarshall %q to ExecCredential: %w", env, error)
```

pkg/token/execCredentialWriter.go
everjing: Changed in new commit d64de9f. Please check it out. Thanks!

pkg/token/execCredentialWriter.go
everjing: Changed per your suggestion in new commit d64de9f. Please check it out. Thanks!

pkg/token/execCredentialWriter.go
```diff
@@ -20,20 +28,63 @@ type execCredentialWriter struct{}
...
+	var execCredential clientauthentication.ExecCredential
+	error := json.Unmarshal([]byte(env), &execCredential)
+	if error != nil {
+		return "", fmt.Errorf("cannot unmarshall %q to ExecCredential: %w", env, error)
+	}
+	switch execCredential.TypeMeta.APIVersion {
+	case "":
+		return apiV1beta1, nil
+	case apiV1:
+		return execCredential.TypeMeta.APIVersion, nil
```
weinong, line 84: fallthrough

everjing: Use case list in switch for the most recent commit 7b8815b. Please check it out. Thanks!

weinong:
Generally looks good to me. but can you please add unit tests?

pkg/token/execCredentialPlugin_test.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/golang/mock/gomock"
 )
 
-func TestExecCredentialPlugin(t *testing.T) {
+func TestExecCredentialWriter(t *testing.T) {
```
weinong, line 15: why is this change needed?

pkg/token/execCredentialWriter.go
```diff
@@ -3,37 +3,91 @@ package token
...
+			},
+			Status: &v1.ExecCredentialStatus{
+				Token:               token.AccessToken,
+				ExpirationTimestamp: &t,
+			},
+		}
+	}
+	var ecCopy interface{} = ec
+	content, _ := json.Marshal(ecCopy)
+	//fmt.Fprintln(os.Stderr, string(content))
```
weinong, line 65: you can use klog with higher verbosity level for logging.

pkg/token/execCredentialWriter.go
```diff
@@ -3,37 +3,91 @@ package token
...
 
+const (
+	apiV1       string = "client.authentication.k8s.io/v1"
+	apiV1beta1  string = "client.authentication.k8s.io/v1beta1"
+	execInfoEnv string = "KUBERNETES_EXEC_INFO"
+)
+
 type ExecCredentialWriter interface {
-	Write(token adal.Token) error
+	Write(token adal.Token, buffer bytes.Buffer) error
```
weinong, line 25:
```suggestion
	Write(token adal.Token, writer io.Writer) error
```
it should be like this so that you pass in os.Stdout from kubelogin and pass in buf.Buffer from unit test

pkg/token/execCredentialWriterAPIVersion_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/Azure/go-autorest/autorest/adal"
+	"k8s.io/client-go/pkg/apis/clientauthentication"
+)
+
+
+func TestExecCredentialWriterAPIVersion() (t *testing.T) {
+	testData1 := []struct {
+		name               string
+		execInfoEnvTest    string
+		stringBufferTest   bytes.Buffer
```
weinong, line 20: you don't need it in the test data, just create a local bytes.Buffer after L59

pkg/token/execCredentialWriterAPIVersion_test.go
```diff
@@ -0,0 +1,69 @@
...
+		expectedAPIVersion string
+	}{
+		{
+			name: "KUBERNETES_EXEC_INFO is empty",
```
weinong, line 24: typical test case should describe the scenario and the expected outcome

pkg/token/execCredentialWriterAPIVersion_test.go
weinong: besides, https://pkg.go.dev/bytes#NewBuffer, 
> In most cases, new(Buffer) (or just declaring a Buffer variable) is sufficient to initialize a Buffer.

pkg/token/execCredentialWriterAPIVersion_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	}
+
+	for _, data := range testData1 {
+		t.Run(data.name, func(t *testing.T) {
+			os.SetEnv("KUBERNETES_EXEC_INFO", testData.execInfoEnvTest)
+			defer os.UnsetEnv("KUBERNETES_EXEC_INFO")
+			execCredentialWriter.Write(adal.Token, data.stringBufferTest)
+			var execCredential clientauthentication.ExecCredential
+			json.Unmarshal([]byte(data.stringBufferTest.String()), &execCredential)
```
weinong, line 62: .Bytes() will return []byte

pkg/token/execCredentialWriterAPIVersion_test.go
everjing: Addressed these two in the most recent commit 244a97c

pkg/token/execCredentialWriterAPIVersion_test.go
everjing: As the test logic for this is relatively easy so I put these in t.Run() func.

weinong:
please remove the extra generated mock files...

pkg/token/execCredentialPlugin_test.go
weinong: @everjing, this file is to test execCredentialPlugin, why is this function changed?

pkg/token/execCredentialWriter.go
```diff
@@ -3,37 +3,92 @@ package token
 //go:generate sh -c "mockgen -destination mock_$GOPACKAGE/execCredentialWriter.go github.com/Azure/kubelogin/pkg/token ExecCredentialWriter"
 
 import (
+	//"bytes"
```
weinong, line 6: please remove unnecessary comments

pkg/token/execCredentialWriter.go
weinong: again, remove unnecessary commented code

pkg/token/execCredentialWriter.go
```diff
@@ -3,37 +3,92 @@ package token
...
+
+func getAPIVersionFromExecInfoEnv() (string, error) {
+	env := os.Getenv(execInfoEnv)
+	if env == "" {
+		return apiV1beta1, nil
+	}
+	var execCredential clientauthentication.ExecCredential
+	error := json.Unmarshal([]byte(env), &execCredential)
+	if error != nil {
+		return "", fmt.Errorf("cannot unmarshall %q to ExecCredential: %w", env, error)
```
weinong, line 84: typo: unmarshal

weinong (changes requested):
so close! please address the comments so that we can merge it

weinong:
you also need to rebase

pkg/token/execCredentialPlugin_test.go
everjing: You are right, this change is not needed and this func name will be reverted in next commit. Thanks!

pkg/token/execCredentialWriter.go
everjing: Done for next commit. Thanks!

pkg/token/execCredentialWriter.go
everjing: Done for next commits. Thanks!

pkg/token/execCredentialWriter.go
everjing: Done in the current commit. Thanks!

pkg/token/execCredentialWriter.go
everjing: Done. Thanks!

pkg/token/execCredentialWriter.go
everjing: Done already. Thanks!

pkg/token/execCredentialWriter.go
everjing: Done already in current commit. Thanks!

pkg/token/execCredentialWriter.go
everjing: Done already in current commit. Thanks!

weinong (approved):
merge it!

PR #145, opened 2022-10-25
Gaojing/devicelogin exit when interactivemode never

everjing:
"Printed out "KUBERNETES_EXEC_INFO" in the code for verification purpose and enabled to return error under condition of "interactiveMode : Never" under devicelogin method. Will do another commit once its logic part is approved. Please check it out. Thanks!


pkg/cmd/token.go
```diff
@@ -3,6 +3,7 @@ package cmd
 import (
 	"github.com/Azure/kubelogin/pkg/token"
```
weinong, line 4: can we revert the change in this file as .... there is no change?

pkg/token/execCredentialPlugin.go
```diff
@@ -28,6 +32,16 @@ type execCredentialPlugin struct {
 }
 
 func New(o *Options) (ExecCredentialPlugin, error) {
+	env := os.Getenv(execInfoEnv)
+	fmt.Fprintln(os.Stderr, os.Getenv(execInfoEnv))
```
weinong, line 36: this line is for debugging, right? you can log it using klog.V(10).Info(), though

pkg/token/execCredentialPlugin.go
```diff
@@ -28,6 +32,16 @@ type execCredentialPlugin struct {
...
+	var execCredential clientauthentication.ExecCredential
+	error := json.Unmarshal([]byte(env), &execCredential)
```
weinong, line 38: a more idiomatic way is
if err := ...; err != nil {
}

pkg/token/execCredentialPlugin.go
```diff
@@ -28,6 +32,16 @@ type execCredentialPlugin struct {
...
+	if error != nil {
+		return nil, fmt.Errorf("cannot convert to ExecCredential: %w", error)
+	}
+	if !execCredential.Spec.Interactive && o.LoginMethod == "DeviceCodeLogin" {
```
weinong, line 42: are you sure it's `"DeviceCodeLogin"` instead of the const?

pkg/token/execCredentialPlugin.go
weinong: besides, you already have `env`. don't need to `os.Getenv()` again?

pkg/cmd/token.go
everjing: Done for next commit.

pkg/token/execCredentialPlugin.go
everjing: Addressed for next commit.

pkg/token/execCredentialPlugin.go
everjing: Changed per comments and please check next commit. Thanks!

pkg/token/execCredentialPlugin.go
everjing: I checked it and you are right. Changed to  DeviceCodeLogin const. Thanks!

pkg/cmd/token.go
```diff
@@ -19,7 +19,6 @@ func NewTokenCmd() *cobra.Command {
 			if err := o.Validate(); err != nil {
 				return err
 			}
-
```
weinong, line 22: i feel this is a unnecessary change?

pkg/token/execCredentialPlugin.go
```diff
@@ -28,6 +32,15 @@ type execCredentialPlugin struct {
 }
 
 func New(o *Options) (ExecCredentialPlugin, error) {
+	env := os.Getenv(execInfoEnv)
+	//fmt.Fprintln(os.Stderr, os.Getenv(execInfoEnv))
```
weinong, line 36: please remove unnecessary commented out code. for debugging purpose, use klog with higher verbosity

pkg/token/execCredentialPlugin.go
```diff
@@ -28,6 +32,15 @@ type execCredentialPlugin struct {
...
+	var execCredential clientauthentication.ExecCredential
+	if error := json.Unmarshal([]byte(env), &execCredential); error != nil {
+		return nil, fmt.Errorf("cannot convert to ExecCredential: %w", error)
+	}
+	if !execCredential.Spec.Interactive && o.LoginMethod == DeviceCodeLogin {
```
weinong, line 41: lets work on #140 first so that you can add appropriate unit test in this PR

pkg/cmd/token.go
everjing: Reverted for next commit.

pkg/token/execCredentialPlugin.go
everjing: Done for next commit. Thanks!

pkg/token/execCredentialPlugin.go
everjing: Added unit test for this commit 8523f02, please check it out. Thanks!

pkg/token/execCredentialPlugin.go
```diff
@@ -29,6 +31,14 @@ type execCredentialPlugin struct {
 }
 
 func New(o *Options) (ExecCredentialPlugin, error) {
+	env := os.Getenv(execInfoEnv)
+	var execCredential clientauthentication.ExecCredential
+	if error := json.Unmarshal([]byte(env), &execCredential); error != nil {
```
weinong, line 36: `error` is a type so generally you don't use the type name as variable name. conventionally, `err` is used

everjing: Updated with err for next commit. Thanks!

weinong (approved):
LGTM

pkg/token/execCredentialPlugin.go
```diff
@@ -29,6 +31,14 @@ type execCredentialPlugin struct {
 }
 
 func New(o *Options) (ExecCredentialPlugin, error) {
+	env := os.Getenv(execInfoEnv)
+	var execCredential clientauthentication.ExecCredential
+	if err := json.Unmarshal([]byte(env), &execCredential); err != nil {
```
yurrriq, line 36: Is it now required to specify `KUBERNETES_EXEC_INFO`? My workflow for v0.0.20 now throws `Error: cannot convert to ExecCredential: unexpected end of JSON input`. It seems perhaps because `env` here is `""` which then can't be unmarshalled into a `clientauthentication.ExecCredential`.

PR #151, opened 2022-11-15
Fix empty "kubernetes_exec_info" env variable broken case

everjing:
It enables case of "kubernetes_exec_info" to work property and pass unit test.


pkg/token/kubernetes_exec_infoIsEmpty_test.go
```diff
@@ -0,0 +1,36 @@
...
+			options: Options{
+				LoginMethod: DeviceCodeLogin,
+				ClientID:    "clientID",
+				ServerID:    "serverID",
+				TenantID:    "tenantID",
+			},
+		},
+	}
+
+	for _, data := range testData {
```
weinong, line 26: i can you can simply update existing unit tests?

everjing: Done in the most recent commit 3c28aca and please check it out. Thanks!

weinong:
LGTM, please squash before we merge

PR #152, opened 2022-11-18
Enable kubectl version lower than 1.22 to work though without triggering exit under devicelogin mode

everjing:
Add checks on kubectl version for lower than 1.22 to issue warnings, provide upgrade options and then work through under devicelogin mode. Passed local unit tests and build successfully locally.

### DIFF
--- a/pkg/token/execCredentialPlugin.go
+++ b/pkg/token/execCredentialPlugin.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/Azure/go-autorest/autorest/adal"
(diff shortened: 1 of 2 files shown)


pkg/token/execCredentialPlugin.go
```diff
@@ -37,7 +38,11 @@ func New(o *Options) (ExecCredentialPlugin, error) {
 		if err := json.Unmarshal([]byte(env), &execCredential); err != nil {
 			return nil, fmt.Errorf("cannot convert to ExecCredential: %w", err)
 		}
-		if !execCredential.Spec.Interactive && o.LoginMethod == DeviceCodeLogin {
+		klog.V(10).Infof("KUBERNETES_EXEC_INFO is: %s", env)
+
+		if strings.Contains(env, `"spec":{}`) {
+			fmt.Fprintln(os.Stderr, "Your current kubectl version is below 1.22.0 and it is highly recommended to use 1.22.0 and above.For upgrading, please refer to https://kubernetes.io/docs/tasks/tools/install-kubectl-linux/")
```
weinong, line 44: i think we can save the message to not annoy people :)

everjing: Done in commit a351ac4. Please check it out. Thanks!

weinong (approved):
LGTM

Issue #173, opened 2022-12-19
kubelogin Installation on Ubuntu 22.04


ddodoo:
I want to ask if there is any support for Linux Os, Ubuntu 22.04 machine

I installed the https://github.com/Azure/kubelogin/releases/download/v0.0.24/kubelogin-linux-amd64.zip release on Ubuntu OS, 22.04

I doesn't seem to work

weinong:
It seems to work for me. How did it not work for you?

```sh
$ uname -a
Linux ubuntu2 5.15.0-1029-azure #36-Ubuntu SMP Mon Dec 5 19:31:08 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
$ ./kubelogin -h
login to azure active directory and populate kubeconfig with AAD tokens

Usage:
  kubelogin [flags]
  kubelogin [command]

Available Commands:
  completion         Generate the autocompletion script for the specified shell
  convert-kubeconfig convert kubeconfig to use exec auth module
  get-token          get AAD token
  help               Help about any command
  remove-tokens      Remove all cached tokens from filesystem

Flags:
  -h, --help          help for kubelogin
      --logtostderr   log to standard error instead of files (default true)
  -v, --v Level       number for the log level verbosity
      --version       version for kubelogin

Use "kubelogin [command] --help" for more information about a command.
$ ./kubelogin --version
kubelogin version
git hash: v0.0.24/bbb308d4255c4a29a17c9bec0f2f89586d084c42
Go version: go1.18.8
Build time: 2022-11-18T22:37:01Z
Platform: linux/amd64
```

ddodoo:
Can you kindly outline how to run the installation?

weinong:
what does `kubelogin -h` show? running `kubelogin` alone yielding nothing is expected. though, there is definitely room to improve to return some useful message

ddodoo:
Hello @weinong 

kubelogin -h does work, thank you for your input

dgroh:
How do I install it on linux?

Tatsinnit:
💡☕️ Hiya, one way I can think of is to install via `azure-cli` route - https://learn.microsoft.com/en-us/cli/azure/install-azure-cli-linux?pivots=apt 

* ☕️ And then using `az aks install-cli` - https://learn.microsoft.com/en-us/cli/azure/aks?view=azure-cli-latest#az-aks-install-cli 

[...]

Issue #89, opened 2022-05-28
Why aren't the common SDK environment variable names used?


tkent:
The environment variables used here are different than those used by the other MSFT SDKs ([dotnet](https://docs.microsoft.com/en-us/dotnet/api/azure.identity.environmentcredential?view=azure-dotnet), [golang](https://docs.microsoft.com/en-us/azure/developer/go/azure-sdk-authentication?tabs=bash), [java](https://docs.microsoft.com/en-us/azure/developer/java/sdk/identity-azure-hosted-auth))

All of them use:

- `AZURE_CLIENT_ID`
- `AZURE_TENANT_ID`
- `AZURE_CLIENT_SECRET`
- `AZURE_USERNAME`
- `AZURE_PASSWORD`

Here, you need to define entirely different variables names like `AAD_SERVICE_PRINCIPAL_CLIENT_ID`. Is there a reason for this inconsistency? Having it means that any execution environment that needs to use both `kubelogin` and an azure SDK must re-define the same values under two different names.

weinong:
nope. there is no reason for that! I think it's at best to support both so that we don't regress existing users. Would you consider to contribute the changes?

tkent:
ATM, I don't have time to contribute patches to this project. I've already accounted for this inconsistency in our pipelines, but believe it would save other MSFT customers time as they move over to `kubelogin` from the currently-defaulted and soon-to-be depreciated `az-login` [aks#2728](https://github.com/Azure/AKS/issues/2728).

weinong:
no problem. we will look into this later this month or next month

GavBurke:
> no problem. we will look into this later this month or next month

Bump, it's now beyond "next month" ;)

weinong:
i'm planning to pick it up next week!

xi2424-npanjwani:
When will this get released?

weinong:
ack. I will release it this week.

PR #174, opened 2022-12-23
Add support of env var convention used by azure sdk

weinong:
* Add support of env var convention used by azure sdk (#89)
* Print help when subcommand is absent (#173)

### DIFF
--- a/pkg/token/options.go
+++ b/pkg/token/options.go
@@ -42,22 +42,37 @@ const (
 	WorkloadIdentityLogin = "workloadidentity"
 	manualTokenLogin      = "manual_token"
 
-	envWorkloadIdentityClientID           = "AZURE_CLIENT_ID"
-	envWorkloadIdentityFederatedTokenFile = "AZURE_FEDERATED_TOKEN_FILE"
-	envWorkloadIdentityAuthorityHost      = "AZURE_AUTHORITY_HOST"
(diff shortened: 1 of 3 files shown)


pkg/token/options.go
```diff
@@ -113,47 +138,75 @@ func (o *Options) UpdateFromEnv() {
 	o.tokenCacheFile = getCacheFileName(o)
 
 	if o.UseAzureRMTerraformEnv {
-		envServicePrincipalClientID = "ARM_CLIENT_ID"
-		envServicePrincipalClientSecret = "ARM_CLIENT_SECRET"
-		envServicePrincipalClientCert = "ARM_CLIENT_CERTIFICATE_PATH"
-		envServicePrincipalClientCertPassword = "ARM_CLIENT_CERTIFICATE_PASSWORD"
-		envTenantID = "ARM_TENANT_ID"
+		if v, ok := os.LookupEnv(terraformClientID); ok {
```
bcho, line 141: suggestion(non-blocking): instead of using a long if-else, maybe we can consider decouple these options into smaller structs, like `AADOptions` / `TerraformOptions`. In those structs, we can make use of env struct libs like this: https://github.com/caarlos0/env to fill in the value. One more benefit is we can decouple the validations into those structs instead of putting in one big `Validate` method.

bcho (approved):
LGTM

PR #177, opened 2023-01-02
Possible enhancement - Add changelog for this repo for automating release tags. ❤️☕️💡

Tatsinnit:
## Intent of the PR

This PR adds the ChangeLog file which will become one file for all the release versions, it kind of follows the keep-the-changelog-standard notion and just one file for all user/PM et. al. to look at to see releases.

* This PR doesn't change the current release behaviour, it just **automating the release tag creation** mechanism, which means repo owners don't need to add release tags and they will be automatically done at the end of the release triggered. In long term if we start and change of binaries with version number then we can add that easily from this release action file.

Screenshot below is from my fork to show it is working. 

* _Things to check:_ Does current mechanism has any dependency of order aka, release tag needs to be created first before release. (as far as I can see its something to do with main branch and how release happens because even from my fork I see the publish looks for release_id route.)

## What is it and advantages it will give & long term benefits for full automation.

As this project will expand and many new contributor might come, the key advantages I see are :

- This encourages more human readable release notes, release automation and more encourages philosophy of “keep all under one roof with structure” strategy in Github project.
- Encourages more peer/ownership eyes, since `changelog.md` PR is separate pull request and key stakeholders can review things before release goes out.
- Avoid ghost releases.
- Avoid commit scavenging for OSS readers or those who are new to the repo.

Thank you so much for reading this and hope this will add value to this repo. ❤️☕️🙏 I hope this helps in long term full automation.

Screenshots:

Screenshot 2023-01-02 at 3 14 45 PM

Screenshot 2023-01-02 at 3 15 16 PM

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -91,28 +91,13 @@ jobs:
         sha256sum kubelogin-linux-amd64.zip > kubelogin-linux-amd64.zip.sha256
         sha256sum kubelogin-linux-arm64.zip > kubelogin-linux-arm64.zip.sha256
 
-    - name: Publish
-      uses: skx/github-action-publish-binaries@master
-      env:
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-      with:
-        args: kubelogin.zip kubelogin-win-amd64.zip kubelogin-darwin-amd64.zip kubelogin-darwin-arm64.zip kubelogin-linux-amd64.zip kubelogin-linux-arm64.zip kubelogin.zip.sha256 kubelogin-win-amd64.zip.sha256 kubelogin-darwin-amd64.zip.sha256 kubelogin-darwin-arm64.zip.sha256 kubelogin-linux-amd64.zip.sha256 kubelogin-linux-arm64.zip.sha256
-
-    - name: Publish to Snap Store
-      uses: snapcore/action-publish@v1
-      env:
-        SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.SNAPSTORE_LOGIN }}
-      with:
-        snap: ${{ steps.snapbuild.outputs.snap }}
-        release: stable
-
     # Read changelog and read versions etc.
     - name: Check version is mentioned in Changelog.md
       id: changelog_reader
       uses: mindsers/changelog-reader-action@v2
       with:
         validation_depth: 10
-        path: './CHANGELOG.md'
+        path: 'CHANGELOG.md'
     # Check if the newest tag already exists
     - name: Check if tag exist
       uses: mukunku/tag-exists-action@5dfe2bf779fe5259360bb10b2041676713dcc8a3 # v1.1.0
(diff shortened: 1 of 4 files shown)


weinong:
i'm not forgetting this PR. Will find some time to review it!

.github/workflows/release.yml
```diff
@@ -95,3 +93,28 @@ jobs:
...
+    - name: Check if tag exist
+      uses: mukunku/tag-exists-action@5dfe2bf779fe5259360bb10b2041676713dcc8a3 # v1.1.0
+      id: check-tag-exists
+      with:
+        tag:  "v${{ steps.changelog_reader.outputs.version }} release" # following the pattern of vx.x.xx release used in this repo
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    # Create Release with artifacts
+    - name: Create Kubelogin Release
+      if: ${{ steps.check-tag-exists.outputs.exists == 'false'}}
```
weinong, line 113: sometimes release may fail when publishing artifacts. would this prevent us from re-running the release?

Tatsinnit: ☕️❤️🥷 Thank you so much for this, I did thought about it, and released we have 2 options:

- **Option 1** - In special case where there are no clear build failure signals and artefacts not published we (dev) could just delete that release tag and restart the workflow.
     - In case of build failure since it runs under same job the workflow will break early and release tag will never be created. 

- **Option 2** - We could separate out these 2 workflow as following:
    -  `Publish` workflow - which is responsible for build and artefact publish, then comes,
    - `Release tag` workflow - which creates the release et. al.

We still be able to automate both the process, Its dev responsibility to first make sure `Publish` runs correctly and artefacts are present **and then** run `Create Release` action.

In the mean time I will see if there is any artefact signal we can check before Crete release tag et. al.

But, please share your thoughts from above and what do you think :) Thank you so much for this discussion.

weinong: Typically it's the last step in publishing artifacts that failed. Code building steps are run before releasing. Option 1 is good for me. Would you mind updating readme.md below please?

Tatsinnit: ☕️ sounds good, I will update and push. Thank you so much for this ❤️🙏

weinong: @Tatsinnit can you rebase? I just merged another change

weinong: btw, in your test release, I don't see published artifacts? is this order right? should release be created before publishing artifacts?

Tatsinnit: ☕️🥷🙏 Thank you so much, sorry was flying (air travelling) hence no replies, now I am land based. Fixed and successful release link from my fork below. Oh in my test runs, I think it must be `release_id` which is getting used for `Publish step` which was not obvious before you asked, I will pass it, also I notices that for `Publish to Snap Store` step some secret is used so from my fork I removed that step to show the success build:  (With `Publish to Snap Store` step as well the build and publish happened but I have shared the runner message below.) 

* https://github.com/Tatsinnit/kubelogin/releases/tag/v0.0.126 
* (All Green) ✅-> https://github.com/Tatsinnit/kubelogin/actions/runs/4056327559/jobs/6980679484 
* (Red with Snap Store ) - I guess due to some secret in use which is not in my Fork  ❌ -> https://github.com/Tatsinnit/kubelogin/actions/runs/4056008295/jobs/6979920553 

I will add readme changes for a scenario where any failure after successful build and publish happened. (As a recommendation for engineers to delete unsuccessful release tag and release notes.)

weinong (approved):
LGTM

PR #179, opened 2023-01-09
Enable CodeQL Analysis.

Tatsinnit:
This PR adds the CodeQL analysis for this repo. This will automate the security checks against this repo. 

Thank you so much in advance, ❤️💡🙏☕️ 

**Extra**: There were some warning but I don't see any logging for secrets and could be associated with this issue for codeql : https://github.com/github/codeql/issues/7992 

`Update`: actually I did see one place where it was logged even with redacted it was not behaving so removing it unless this logger is necessary.

### DIFF
--- a/pkg/token/execCredentialPlugin.go
+++ b/pkg/token/execCredentialPlugin.go
@@ -29,7 +29,6 @@ type execCredentialPlugin struct {
 }
 
 func New(o *Options) (ExecCredentialPlugin, error) {
-	klog.V(10).Info(o)
 	provider, err := newTokenProvider(o)
 	if err != nil {
 		return nil, err
--- a/pkg/token/execCredentialPlugin.go
+++ b/pkg/token/execCredentialPlugin.go
@@ -29,7 +29,10 @@ type execCredentialPlugin struct {
 }
 
 func New(o *Options) (ExecCredentialPlugin, error) {
-	klog.V(10).Info(o.String())
+
+	logginOptionsObject := marshalOptionsForLogging(o)
+
+	klog.V(10).Info(logginOptionsObject)
 	provider, err := newTokenProvider(o)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


pkg/token/execCredentialPlugin.go
```diff
@@ -29,7 +29,6 @@ type execCredentialPlugin struct {
 }
 
 func New(o *Options) (ExecCredentialPlugin, error) {
-	klog.V(10).Info(o)
```
weinong, line 32: why is this line removed?

weinong: oh. is it flagged by CodeQL? I wonder which part triggered it. It'd be useful to log the options, though.

Tatsinnit: ❤️☕️💡💡 Thank you so much for this 🙏, indeed, I did some more thinking and will share here for thoughts and ideas - I feel like we have 2 options. (This is from few hours of thinking I was doing while COdeQL threw this error)

Prima Facie, it's like a vanilla suspision from codeQL that - Hey this will print Creds in plain text. **But** that could very well be use case scenario so please see what I thought and might be correct me or validate.

Change above is very basic and removes the issue.

### Option-1 - Might be log is needed (for debuggability) - in that case no need to add codeQL

* Although, logging crews in plain-text could be not-preferred but in case of presumption of local host gets hijacked, hacker will have access to `kubeconfig` as well so anyone can do more destructive action then just this.

### Option 2 - If we add CodeQL - we have 2 possible solution - short term and might be long term

* Short term solution to remove the klog for object which had credentials and password. CodQL is happy.
* One long term solution - revisit the `Options` object to refactor for all credentials as part of separate sub-object,  even with redact the password or `encrypt`. 🥷🥷

Thanks heaps for this. 💡❤️

weinong: what if you change it to `klog.V(10).Info(o.String())`. I don't believe this log will print the secret as it implements the [stringer](https://github.com/Azure/kubelogin/blob/master/pkg/token/options.go#L215) interface and will not return the credential. but I could be mistaken by `Info` by `klog`

Tatsinnit: ❤️☕️ It almost feel like that codeQL scanning is very rudimentary. I have fixed it by creating a logging only `struct` -> `KlogsLoggingPurposeOptions` I have a feeling that codeQl analyser is not deep analyser, it basically checks the type and if its string for fields like -> secrets, crews, password - it throws error. 

Weird thing: Just as test I also tried with `crypt` hash and it still dint liked it because of `Clear-text logging of sensitive information` rule, I feel like it thinks as if `klog.Info` will log in plain site.

But with this solution the new struct for logging only form struct without password, secret or credit fields.

* CodeQL if all green now 🎊 and passing. 

Screenshot 2023-01-11 at 5 50 24 PM


Please take a look.

weinong (approved):
LGTM

PR #182, opened 2023-01-11
build: add support for Ubuntu snap package

### LINKED ISSUE #167: Provide apt-get and/or snap package
This also makes the package auto-discoverable on Linux command line as the default shell will suggest apt and snap packages containing a command that you tried to run.

Exodus:
This closes #167 

Adding support for snap package.

This does as much as possible to align with the way versioning is set in the project, using `git describe --tags` for the snap package itself as the Makefile does.

Additionally, the kubelogin name must be reserved in snapcraft, which I can do if this PR is accepted as well as set up automatic builds for there.

@weinong

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,25 @@
+name: kubelogin # you probably want to 'snapcraft register <name>'
+base: core22 # the base snap is the execution environment for this snap
+adopt-info: kubelogin
+summary: Azure kubelogin # 79 char long summary
+description: |
+  A Kubernetes credential (exec) plugin implementing azure authentication.
+
+grade: stable # must be 'stable' to release into candidate/stable channels
+confinement: strict # use 'strict' once you have the right plugs and slots
+apps:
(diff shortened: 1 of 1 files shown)


snap/snapcraft.yaml
```diff
@@ -0,0 +1,25 @@
+name: kubelogin # you probably want to 'snapcraft register <name>'
```
weinong, line 1: should this be azure-kubelogin to avoid name collision?

Exodus: No, just kubelogin, as the idea is this is the official package, thus named 'kubelogin'. There is currently no other package named kubelogin, this one would be it.

weinong (approved):
Thanks!

Exodus:
I have uploaded the build to snap now, kubelogin is available. I'd like to follow up with two items:
- Enable automatic builds in snapcraft
- Add the code owners here so they have access in snapcraft (@weinong I have added you since you mentioned you added a snapcraft account in #167)

Issue #188, opened 2023-02-08
Missing version information of the binary.


szymonos:
There is no version information when running command `kubelogin --version`. Returned result:

``` text
kubelogin version 
git hash: master/b46c5fe18e6f295173c5062b7f30ad623844faef
Go version: go1.19.5
Build time: 2023-02-08T00:52:24Z
Platform: linux/amd64
```

In previous releases there was information about kubelogin version in the 1st line.
I have a script that automatically updates version based on the latest release in GitHub repository and right now it basically downloads it every time, because I cannot determine the currently installed version.

Tatsinnit:
Thank you so much for reporting this, @szymonos, really appreciate it. It is fixed under following PR with details as to what was happening #189 ❤️❤️🙏

Thank you and cc: @weinong + @peterbom

szymonos:
Thank you for the fast response :D. Works perfectly now :). Yep, it was second line ;),

PR #189, opened 2023-02-09
Fix/release tagging

Tatsinnit:
This PR fixes : #188 , thank you so much for reporting this.

### TLDR:

The issue was the missing automation for `pull tags` to provide all latest tags before build for the `version information` to follow.

### Little extra:

Previously we were creating the tags after `build` and also since we are now automating everything, I completely missed out this part of the code which was handling the version and not we are making sure all tags are there so the hash with the latest tag matches.

Screenshot 2023-02-09 at 1 06 49 PM

As a result here is what I get Before and After:

### Before fix

Screenshot 2023-02-09 at 11 51 03 AM

### After Fix

Screenshot 2023-02-09 at 1 10 47 PM

Thank you so much for this, really appreciate it. ❤️☕️🙏

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,6 +15,33 @@ jobs:
         go-version: 1.19
       id: go
 
+    # Read changelog and read versions etc.
+    - name: Check version is mentioned in Changelog.md
+      id: changelog_reader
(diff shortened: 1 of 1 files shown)


weinong (approved):
LGTM

Issue #197, opened 2023-02-15
Vulnerability: Authorization Bypass Through User-Controlled Key


cloud-operations:
Found below vulnerability in 0.0.26 kubelogin version:

https://avd.aquasec.com/nvd/2022/cve-2022-1996/

Please take a look

Thanks

Tatsinnit:
💡☕️ Thank you @cloud-operations for this. 

**Quick build up question please**: from the above link are you suggesting that the indirect usage of `emicklei/go-restful ` or did you find anything specific for the `kubelogin`?

cloud-operations:
Thanks for quick turnaround. When I install v.0.0.26 version of kubelogjn, it comes with v3.8.0 version of emicklei/go-restful which contains the subjected vulnerability. Looks like it is fixed in v.3.10.1 version. However, I am not able to upgrade go-restful specifically as it has to come through a new/fixed version of kubelogin.

weinong:
fixed in v0.0.27

PR #198, opened 2023-02-16
update go-restful mod.

Tatsinnit:
Thank you for taking time and reviewing this PR, this PR could cater - #197  - but this is an indirect usage and our in repo has  dependent security bot has **not** warned/complained us about it. ☕️❤️🙏🚀

Although the packages cahngelog don't mention any change of funct signature or breaking change here: https://github.com/emicklei/go-restful/blob/v3/CHANGES.md#v3101---2022-11-19 - safe side a gentle ping to @weinong - in case this innocent change end up with some breaking functs and calls.

Thanks heaps,

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -26,7 +26,7 @@ require (
 	github.com/Azure/go-autorest/logger v0.2.1 // indirect
 	github.com/Azure/go-autorest/tracing v0.6.0 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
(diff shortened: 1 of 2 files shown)


Tatsinnit:
💡 🥷 🙏 Note to self - Might be - I should check all the direct dependencies and update that which will bring underlying indirect dependencies to stable. (Will try this sometime this week)

weinong (approved):
LGTM

PR #202, opened 2023-02-22
document the default device code login doesn't work with conditional …

weinong:
document the default device code login doesn't work with conditional access policy

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ kubectl get no
 
 > Note: although device code flow is the default login mode, it doesn't work when Conditional Access policy is configured on AAD tenant. Use web browser flow instead.
 
-> Note: If you are using kubeconfig from AKS AADv1 clusters, `convert-kubeconfig` command will automatically add `--legacy` flag so that `audience` claim will have `spn:` prefix.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -71,7 +71,9 @@ kubelogin convert-kubeconfig
 kubectl get no
 ```
 
-If you are using kubeconfig from AKS AADv1 clusters, `convert-kubeconfig` command will automatically add `--legacy` flag so that `audience` claim will have `spn:` prefix.
+> Note: although device code flow is the default login mode, it doesn't work when Conditional Access policy is configured on AAD tenant. Use web browser flow instead.
+
+> Note: If you are using kubeconfig from AKS AADv1 clusters, `convert-kubeconfig` command will automatically add `--legacy` flag so that `audience` claim will have `spn:` prefix.
```
bcho, line 76: nit: use `if` to align with the `although` case above?

bcho, line 76:
```suggestion
> Note: if you are using kubeconfig from AKS AADv1 clusters, `convert-kubeconfig` command will automatically add `--legacy` flag so that `audience` claim will have `spn:` prefix.
```

bcho (approved):
LGTM

PR #210, opened 2023-02-23
ci: remove snapstore publish step from push action

bcho:
The snap build seems broken and preventing depedantbot's PR builds.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -57,3 +57,7 @@ jobs:
         GOOS: windows
         GOARCH: arm64
       run: make
+
+    - name: Snap Build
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -57,18 +57,3 @@ jobs:
         GOOS: windows
         GOARCH: arm64
       run: make
-
-    - name: Snap Build
-      uses: snapcore/action-build@v1
-      id: snapbuild
```
weinong, line 63: keep the build?

bcho: done, ptal

weinong (approved):
LGTM

PR #229, opened 2023-03-14
Enable Code Cov for this repo.

Tatsinnit:
Hiya, This is Part 0/1 of CodeCov enablement which will enable codecov on build and a sample way it will trigger is this:

❤️☕️💡🙏 Please take a look and if it looks good, this will also help us to increase as part 2 onwards and see the improvement from baseline percentage et. al. 

Thanks heaps and fyi: @weinong . ☕️🙏

* **Sample only PR from my Fork** https://github.com/Tatsinnit/kubelogin/pull/5

Screenshot what this change will achieve:

Screenshot 2023-03-15 at 10 15 20 AM

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -61,3 +61,6 @@ jobs:
     - name: Snap Build
       uses: snapcore/action-build@v1
       id: snapbuild
+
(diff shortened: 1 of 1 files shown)


weinong (approved):
LGTM

PR #238, opened 2023-03-24
Feature/addtests

Tatsinnit:
This PR intends to add more test with intention of catering low hanging fruits of capturing failure case tests and hence increasing the test coverage to almost 50% for this repo.

I have used direct calls and no mocks are involved so the correct logic can be test.

Thank yo use much in advance. ❤️🙏 Kind ping as fyi @weinong 

^Tats

Screenshot 2023-03-24 at 6 26 52 PM

### DIFF
--- a/pkg/token/azurecli_test.go
+++ b/pkg/token/azurecli_test.go
@@ -1,0 +1,33 @@
+package token
+
+import (
+	"strings"
+	"testing"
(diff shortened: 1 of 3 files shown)


weinong (approved):
Apologize for missing this PR. LGTM.

PR #242, opened 2023-04-06
Document scoop installation option

goostleek:
It's possible to install `kubelogin` using [scoop](https://github.com/ScoopInstaller/Scoop#scoop) installer. The [manifest file](https://github.com/ScoopInstaller/Main/blob/master/bucket/azure-kubelogin.json) is automatically updated when a [new release](https://github.com/Azure/kubelogin/releases) comes out.


weinong (approved):
fancy! thanks for the update!

PR #247, opened 2023-04-07
update readme and docs

weinong:


docs/book/src/concepts/login-modes/sp.md
```diff
@@ -77,5 +77,5 @@ kubectl get nodes
 ## Restrictions
 
 - on AKS, it will only work with managed AAD
-- the service principal can be member of maximum 250 AAD groups
+- the service principal can be member of maximum 200 AAD groups
```
bcho, line 80: question(non-blocking): I wonder do we have any official aks/aad doc that tell such restrictions? Maybe we can back link to that doc in case of future update

bcho (approved):
LGTM

docs/book/src/concepts/login-modes/sp.md
Tatsinnit: 💡 🥷 Could be --> It is probably due to this 200 single user to directories limit: https://learn.microsoft.com/en-us/azure/active-directory/enterprise-users/directory-service-limits-restrictions

Screenshot 2023-04-10 at 5 37 49 PM

weinong: yes. feel free to raise a PR to add this reference: https://learn.microsoft.com/en-us/azure/active-directory/hybrid/how-to-connect-fed-group-claims

PR #260, opened 2023-04-21
add --context support in convert subcommand

### LINKED ISSUE #87: Support converting a single cluster in a kubeconfig file
It would be nice if kubelogin supported converting a single cluster in a kubeconfig file, and let the other cluster configurations be as-is to reduce the blast radius.

weinong:
it allows to convert only the selected context rather than the whole kubeconfig

fixes #87

### DIFF
--- a/pkg/converter/convert.go
+++ b/pkg/converter/convert.go
@@ -149,7 +149,17 @@ func Convert(o Options, pathOptions *clientcmd.PathOptions) error {
 		return fmt.Errorf("unable to load kubeconfig: %s", err)
 	}
 
-	for _, authInfo := range config.AuthInfos {
+	targetAuthInfo := ""
+
+	if o.context != "" && config.Contexts[o.context] != nil {
+		targetAuthInfo = config.Contexts[o.context].AuthInfo
+	}
+
+	for name, authInfo := range config.AuthInfos {
+
+		if targetAuthInfo != "" && name != targetAuthInfo {
+			continue
+		}
 
 		//  is it legacy aad auth or is it exec using kubelogin?
 		if !isExecUsingkubelogin(authInfo) && !isLegacyAzureAuth(authInfo) {
--- a/pkg/converter/convert_test.go
+++ b/pkg/converter/convert_test.go
@@ -14,7 +14,8 @@ import (
 
 func TestConvert(t *testing.T) {
 	const (
-		clusterName        = "aks"
+		clusterName1       = "aks1"
+		clusterName2       = "aks2"
 		envName            = "foo"
 		serverID           = "serverID"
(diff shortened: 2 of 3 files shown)


Tatsinnit (approved):
💡☕️🎉 looks good to me.

pkg/converter/convert.go
```diff
@@ -149,7 +149,17 @@ func Convert(o Options, pathOptions *clientcmd.PathOptions) error {
...
+	targetAuthInfo := ""
+
+	if o.context != "" && config.Contexts[o.context] != nil {
+		targetAuthInfo = config.Contexts[o.context].AuthInfo
+	}
+
+	for name, authInfo := range config.AuthInfos {
+
+		if targetAuthInfo != "" && name != targetAuthInfo {
+			continue
```
bcho, line 161: question: what's the expected behavior when there is no auth info matches the passed context name? Should we fail the call?

pkg/converter/convert_test.go
```diff
@@ -1042,6 +1044,26 @@ func TestConvert(t *testing.T) {
...
+			authProviderConfig: map[string]string{
+				cfgEnvironment: envName,
+				cfgApiserverID: serverID,
+				cfgClientID:    clientID,
+				cfgTenantID:    tenantID,
+				cfgConfigMode:  "0",
+			},
+			overrideFlags: map[string]string{
+				flagLoginMethod: token.MSILogin,
+				"context":       clusterName1,
```
bcho, line 1059: nit: this flag name is repeating multiple times, consider adding a `flagContext` variable for this usage?

pkg/converter/convert_test.go
```diff
@@ -1042,6 +1044,26 @@ func TestConvert(t *testing.T) {
 			},
 			command: execName,
 		},
+		{
+			name:        "convert with context specified, auth info not specified by the context should not be changed",
```
bcho, line 1048: maybe we should also add a test case for context specified, auth info not found case to ensure the behavior

pkg/converter/convert.go
weinong: it'd be no ops. meaning the kubeconfig will not be updated at all

pkg/converter/convert_test.go
weinong: i think what you are suggesting is the essence of this test case? for any unmatched auth info, they should remain as is

bcho (approved):
LGTM. My remaining question is on what should we show to the user if there is no matched context. This is a common case if there is typo in the `--context` value. Silently omitting this might cause confusion in some cases. But I am fine to improve this part in later change.

PR #272, opened 2023-05-02
added verbose logging in convert-kubeconfig

weinong:


pkg/converter/convert.go
```diff
@@ -168,10 +184,15 @@ func Convert(o Options, pathOptions *clientcmd.PathOptions) error {
 			continue
 		}
 
+		klog.V(5).Infof("context: %q", name)
+
 		//  is it legacy aad auth or is it exec using kubelogin?
 		if !isExecUsingkubelogin(authInfo) && !isLegacyAzureAuth(authInfo) {
```
bcho, line 190: do we want a log in this continue case?

bcho (commented):
LGTM with a minor comment

pkg/converter/convert.go
weinong: nah. the L155 has enough info to tell whether the context is specified. anything not printed here will mean that's continued. besides, i don't think we need to log unmatched context

bcho (approved):

PR #282, opened 2023-05-18
Adding installHint field to kubeconfigs that have been converted to the exec format

cirvine-MSFT:
InstallHint field tells users how to install the binary in the command field if it isn't present in their PATH.
We need to add this field to any kubeconfig that gets converted since it can be transported to another machine/environment that does not have kubelogin installed.

### DIFF
--- a/pkg/converter/convert.go
+++ b/pkg/converter/convert.go
@@ -200,8 +200,8 @@ func Convert(o Options, pathOptions *clientcmd.PathOptions) error {
 			Args: []string{
 				getTokenCommand,
 			},
-			APIVersion: execAPIVersion,
-			InstallHint: execInstallHint
+			APIVersion:  execAPIVersion,
+			InstallHint: execInstallHint,
 		}
 
 		exec.Args = append(exec.Args, argLoginMethod, o.TokenOptions.LoginMethod)
--- a/pkg/converter/convert.go
+++ b/pkg/converter/convert.go
@@ -51,7 +51,6 @@ const (
 	flagAuthorityHost      = "authority-host"
 	flagFederatedTokenFile = "federated-token-file"
 	flagTokenCacheDir      = "token-cache-dir"
-	flagExecInstallHint    = "install-hint"
 
 	execName        = "kubelogin"
 	getTokenCommand = "get-token"
@@ -203,10 +202,6 @@ func Convert(o Options, pathOptions *clientcmd.PathOptions) error {
 			},
 			APIVersion:  execAPIVersion,
 			InstallHint: execInstallHint,
-		}
-
-		if o.execInstallHint != "" {
-			exec.InstallHint = o.execInstallHint
 		}
 
 		exec.Args = append(exec.Args, argLoginMethod, o.TokenOptions.LoginMethod)
--- a/pkg/converter/convert.go
+++ b/pkg/converter/convert.go
@@ -208,6 +208,11 @@ func Convert(o Options, pathOptions *clientcmd.PathOptions) error {
 			InstallHint: execInstallHint,
 		}
 
+		// Preserve any existing install hint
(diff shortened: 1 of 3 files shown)


pkg/converter/convert.go
```diff
@@ -199,7 +200,8 @@ func Convert(o Options, pathOptions *clientcmd.PathOptions) error {
 			Args: []string{
 				getTokenCommand,
 			},
-			APIVersion: execAPIVersion,
+			APIVersion:  execAPIVersion,
+			InstallHint: execInstallHint,
```
weinong, line 204: let's use the install hint from input, when provided, before falling back to the hint here

cirvine-MSFT: Thanks for the suggestion @weinong. I think my latest push contains the changes you wanted. PTAL.

pkg/converter/convert.go
```diff
@@ -51,10 +51,12 @@ const (
 	flagAuthorityHost      = "authority-host"
 	flagFederatedTokenFile = "federated-token-file"
 	flagTokenCacheDir      = "token-cache-dir"
+	flagExecInstallHint    = "install-hint"
 
 	execName        = "kubelogin"
 	getTokenCommand = "get-token"
 	execAPIVersion  = "client.authentication.k8s.io/v1beta1"
+	execInstallHint = "Install kubelogin for use with kubectl by running az aks install-cli"
```
weinong, line 59:
```suggestion
	execInstallHint = "Install kubelogin to connect to Azure AD enabled Kubernetes clusters. Go to https://azure.github.io/kubelogin for details"
```

weinong: i tend not to put aks context here as this tool can be used for any Azure AD k8s, even on AWS. :)

pkg/converter/options.go
```diff
@@ -33,6 +34,7 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 	}
 	fs.StringVar(&o.context, flagContext, "", "The name of the kubeconfig context to use")
 	fs.StringVar(&o.azureConfigDir, flagAzureConfigDir, "", "Azure CLI config path")
+	fs.StringVar(&o.execInstallHint, flagExecInstallHint, "", "The install hint for the kubelogin command when converting to exec format.")
```
weinong, line 37: actually, I didn't mean to add an option for the hint. What i meant is, when the kubeconfig to be converted is already of exec format and has install hint, we should preserve it during conversion. For instance
```
az aks get-credentials... 
# the downloaded kubeconfig is in exec format and has aks specific install hint
kubelogin convert-kubeconfig -l msi
# the output should keep the same hint as before
```

pkg/converter/convert.go
cirvine-MSFT: Makes sense and thanks for that feedback. I did not commit your suggestion as is, but instead aligned with the install hint we are adding to AKS. Let me know if you didn't want that.

pkg/converter/options.go
cirvine-MSFT: Makes sense. Reverted that change.

pkg/converter/convert.go
```diff
@@ -55,6 +55,11 @@ const (
 	execName        = "kubelogin"
 	getTokenCommand = "get-token"
 	execAPIVersion  = "client.authentication.k8s.io/v1beta1"
+	execInstallHint = `
+kubelogin is not installed which is required to connect to AAD enabled cluster.
+
+To learn more, please go to https://aka.ms/aks/kubelogin
```
weinong, line 61: let's use https://azure.github.io/kubelogin/

weinong, line 61:
```suggestion
To learn more, please go to https://azure.github.io/kubelogin/
```

weinong (approved):
LGTM

PR #298, opened 2023-06-20
docs: Use asdf-plugins index instead of hard coded repo

sechmann:
This makes it less likely these docs will break in the event that this plugin repository moves.


weinong (approved):
LGTM

PR #299, opened 2023-06-29
Add chocolatey installation instructions

moredatapls:
Adding chocolatey installation instructions. They would've helped me as I first confused the package with https://community.chocolatey.org/packages/kubelogin (https://github.com/int128/kubelogin), a different "kubelogin" plugin.

The package seems to not be maintained by MS directly, but neither is the Scoop one, so I added the disclaimer here as well.


weinong (approved):
LGTM. Thanks!

Issue #102, opened 2022-07-14
The kubectl commands become significant slower when using kubelogin


JiayangZhou:
In AKS, we started to get warnings such as below:
![image](https://user-images.githubusercontent.com/47085823/178996615-32899a92-9aab-484d-881c-92adb65892ce.png)

So we adopted kubelogin by running the below commands: `kubelogin convert-kubeconfig -l azurecli`, which is non-interactively login to AKS. And then every time we run a kubectl command, we can feel a delayed response time. The issue is gone if do the normal way of refreshing AKS token via `az aks get-credentials`.

The kubeconfig file looks something like this when the issue is present:

```
- name: clusterUser_clustername
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      args:
      - get-token
      - --server-id
      - XXXXX
      - --login
      - azurecli
      command: kubelogin
      env: null
      provideClusterInfo: false
```

This has been reported by many of our developers, but everybody uses the same AKS cluster, we will also initiate an Azure support ticket and see if it has something to do on cluster level, and also try in different clusters. But feel this is more related to kubelogin.

JiayangZhou:
Ran a test with this bash script:
```
start_time=$(date +%s)
# perform a task
kubectl get ns
kubectl get pods
kubectl get pv
kubectl get ingress -A
kubectl config view
kubectl get secret
kubectl get jobs
kubectl get deployment

end_time=$(date +%s)

# elapsed time with second resolution
elapsed=$(( end_time - start_time ))
echo "==================================="
echo "The time elapsed is:" $elapsed"s"
```

and the result is:
<html>
<body>


dev aks + non-interactive | 12 | 33 | 11 | 10
-- | -- | -- | -- | --
dev aks + interactive | 9 | 7 | 9 | 8


</body>
</html>

weinong:
@JiayangZhou kubelogin caches token from azurecli, and the first invocation may be slower than the rest. In my testing below, it seems reasonable, albeit my cluster is very empty.

```
#!/bin/bash

function run () {
start_time=$(date +%s)
$1 > /dev/null 2>&1
end_time=$(date +%s)
elapsed=$(( end_time - start_time ))
echo $1
echo "The time elapsed is:" $elapsed"s"
echo "==================================="
}


# clean up the cache
kubelogin remove-tokens

# perform a task
run "kubectl get ns"
run "kubectl get pods"
run "kubectl get pv"
run "kubectl get ingress -A"
[...]

PR #306, opened 2023-07-18
Upgrade klog from v1 to v2

peterbom:
Addresses #102.

`v2` of `klog` no longer calls `user.Current()` which hits the SAM RPC endpoint that can cause slow startup on Windows.


Tatsinnit (approved):
❤️☕️💥

weinong (approved):
Awesome find!

PR #319, opened 2023-08-15
Add PoP token support to interactive+spn get-token/convert-kubeconfig flows

rharpavat:
Adds support for getting a [Proof-of-Possession (PoP) token](https://learn.microsoft.com/en-us/entra/msal/dotnet/advanced/proof-of-possession-tokens) when running `kubelogin get-token` with `--login interactive` mode.

Adds 2 new flags to `kubelogin get-token` and `kubelogin convert-kubeconfig`:

1. `--pop-enabled`: boolean flag indicating whether or not to request a PoP token
2. `--pop-claims`: string slice formatted as a comma-separated list of key/value pairs, eg. `"u=host,key2=val2"`

Both flags must be provided together in order to use the PoP token flow; providing one but not the other will result in an error during flag validation. PoP claims must contain the `u` claim containing the ARM ID of the cluster (host) to request the PoP token for, in the format `"u=<ARM_ID>"`.


pkg/pop/poptoken.go
```diff
@@ -0,0 +1,184 @@
...
+	payload := fmt.Sprintf(`{"at":"%s","ts":%d,"u":"%s","cnf":{"jwk":%s},"nonce":"%s"}`, accessToken, ts, as.Host, as.PoPKey.JWK(), nonce)
+	payloadB64 := base64.RawURLEncoding.EncodeToString([]byte(payload))
+	h256 := sha256.Sum256([]byte(headerB64 + "." + payloadB64))
+	signature, err := as.PoPKey.Sign(h256[:])
+	if err != nil {
+		return "", err
+	}
+	signatureB64 := base64.RawURLEncoding.EncodeToString(signature)
+
+	return headerB64 + "." + payloadB64 + "." + signatureB64, nil
```
julienstroheker, line 56: Can we create a struct for this ?

julienstroheker: instead of manipulating string directly via `fmt.Sprintf()` in the function we can for sure get a representation of what is a pop token via a struct, something like :
```
type PoPAccessToken struct {
   Header header
   Payload payload
   Signature signature
}

type header struct {
  typ string
  alg string
  kid string
}

func (h *header) ToString() string {
  return fmt.Sprintf({"typ":"%s","alg":"%s","kid":"%s"}, h.typ, h.alg, h.kid)
}

func (h *header) ToBase64() string {
  base64.RawURLEncoding.EncodeToString([]byte(h.ToString()))
}
```

pkg/pop/poptoken.go
```diff
@@ -0,0 +1,184 @@
...
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/google/uuid"
+)
+
+const popTokenType = "pop"
+
+type PopAuthenticationScheme struct {
```
julienstroheker, line 24: Eveything about the authnscheme from msal could be moved to a dedicated file in this same pkg `authnscheme.go`. 

`poptoken.go` should only be the interface and the implementation of it

pkg/pop/poptokenutils.go
```diff
@@ -0,0 +1,73 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/confidential"
+	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/public"
+)
+
+func AcquirePoPTokenInteractive(
```
julienstroheker, line 14: Add description on top of each function to describe what it is

pkg/pop/poptokenutils.go
```diff
@@ -0,0 +1,73 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+package pop
```
julienstroheker, line 4: Let's rename this file from `poptokenutils.go` to `msal.go` ?

pkg/pop/poptokenutils.go
```diff
@@ -0,0 +1,73 @@
...
+	scopes []string,
+	authority,
+	clientId string,
+) (string, int64, error) {
+	client, err := public.New(clientId, public.WithAuthority(authority))
+	if err != nil {
+		return "", -1, fmt.Errorf("unable to create public client: %w", err)
+	}
+	result, err := client.AcquireTokenInteractive(
+		context.Background(),
```
julienstroheker, line 25: It is usually a good practice to get this context from the caller. This can be a parameter of the function

pkg/token/options.go
```diff
@@ -118,6 +120,8 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&o.IsLegacy, "legacy", o.IsLegacy, "set to true to get token with 'spn:' prefix in audience claim")
 	fs.BoolVar(&o.UseAzureRMTerraformEnv, "use-azurerm-env-vars", o.UseAzureRMTerraformEnv,
 		"Use environment variable names of Terraform Azure Provider (ARM_CLIENT_ID, ARM_CLIENT_SECRET, ARM_CLIENT_CERTIFICATE_PATH, ARM_CLIENT_CERTIFICATE_PASSWORD, ARM_TENANT_ID)")
+	fs.BoolVar(&o.IsPoPTokenEnabled, "pop-enabled", o.IsPoPTokenEnabled, "set to true to use a PoP token for authentication or false to use a regular bearer token")
+	fs.StringVar(&o.PoPTokenClaims, "pop-claims", o.PoPTokenClaims, "contains a comma-separated list of claims to attach to the pop token in the format `key=val,key2=val2`. At minimum, specify the ARM ID of the connected cluster as `u=ARM_ID`")
```
julienstroheker, line 124: I know that we want to be really specific to our scenario but I would avoid to mention `connected cluster`. Can just be `At minimum, specify the ARM ID of the cluster as `u=ARM_ID`"`

pkg/token/options.go
```diff
@@ -236,3 +250,29 @@ func getCacheFileName(o *Options) string {
...
+		}
+		key := strings.TrimSpace(claimPair[0])
+		val := strings.TrimSpace(claimPair[1])
+		if key == "" || val == "" {
+			return nil, fmt.Errorf("error parsing PoP token claims. Ensure the claims are formatted as `key=value` with no extra whitespace")
+		}
+		claimsMap[key] = val
+	}
+	if claimsMap["u"] == "" {
+		return nil, fmt.Errorf("required u-claim not provided for PoP token flow. Please provide the ARM ID of the connected cluster in the format `u=<ARM_ID>`")
```
julienstroheker, line 275: same remark, we should just mention `cluster` not `connected cluster`

pkg/token/options_test.go
```diff
@@ -149,9 +166,53 @@ func TestOptionsWithEnvVars(t *testing.T) {
...
 			o.UpdateFromEnv()
-			if o != tc.expected {
+			if !cmp.Equal(o, tc.expected, cmp.AllowUnexported(Options{})) {
 				t.Fatalf("expected option: %+v, got %+v", tc.expected, o)
 			}
 		})
 	}
 }
+
+func TestParsePoPClaims(t *testing.T) {
```
julienstroheker, line 176: This can be easly converter to a TDT model to avoid duplication

```

func Test(t *testing.T) {
	testCases := []struct {
		desc	string
		
	}{
		{
			desc: "",
			
		},
	}
	for _, tC := range testCases {
		t.Run(tC.desc, func(t *testing.T) {
			
		})
	}
}
```

pkg/token/options_test.go
```diff
@@ -149,9 +166,53 @@ func TestOptionsWithEnvVars(t *testing.T) {
...
+
+	t.Run("pop-claim parsing should fail if claims are malformed", func(t *testing.T) {
+		popClaims := "claim1=  "
+		if _, err := parsePopClaims(popClaims); err == nil || !strings.Contains(err.Error(), "Ensure the claims are formatted as `key=value`") {
+			t.Fatalf("parsing pop claims should return error if claims are malformed. got: %s", err)
+		}
+	})
+
+	t.Run("pop-claim parsing should fail if u-claim is not provided", func(t *testing.T) {
+		popClaims := "claim1=val1, claim2=val2"
```
julienstroheker, line 206: `"claim1=val1, claim2=val2"` CLI will parse this option without the space right ? `"claim1=val1,claim2=val2"`

pkg/token/options_test.go
```diff
@@ -149,9 +166,53 @@ func TestOptionsWithEnvVars(t *testing.T) {
...
+		if _, err := parsePopClaims(popClaims); err == nil || !strings.Contains(err.Error(), "required u-claim not provided") {
+			t.Fatalf("parsing pop claims should return error if u-claim is not provided. got: %s", err)
+		}
+	})
+
+	t.Run("pop-claim parsing should succeed with u-claim and additional claims", func(t *testing.T) {
+		popClaims := "u=val1, claim2=val2, claim3=val3"
```
julienstroheker, line 213: Let's test with `1=2,3=4` & `,,,,,,,,,,,`

pkg/token/interactive.go
```diff
@@ -51,32 +54,54 @@ func (p *InteractiveToken) Token() (adal.Token, error) {
...
-	if err != nil {
-		return emptyToken, fmt.Errorf("expected an empty error but received: %w", err)
-	}
-	if interactiveToken.Token == "" {
-		return emptyToken, errors.New("did not receive a token")
+	var token string
+	var expirationTimeUnix int64
+	var err error
+
+	if p.popClaims == nil || len(p.popClaims) == 0 {
```
julienstroheker, line 63: So at that part of the code, you relying on the fact that claims should not be empty if the enabled flag is passed along with the u claim right ?

rharpavat: Yup, the initial validation in `options.go` ensures both flags are either passed together or not at all. Then the check when getting the token provider in `provider.go` ensures that if the pop-enabled flag is present, the pop-claims parse correctly and u-claim is present before providing pop claims to the token provider.

pkg/token/options_test.go
rharpavat: Yeah the pop token claim parsing trims spaces from each key/value

pkg/pop/poptokenutils.go
rharpavat: Renamed

pkg/pop/poptokenutils.go
rharpavat: Updated to pass context from caller

pkg/token/options.go
rharpavat: Updated to just say 'cluster'

pkg/token/options.go
rharpavat: Updated to just say 'cluster'

pkg/token/options_test.go
rharpavat: Converted to this model

pkg/token/options_test.go
rharpavat: Added these test cases

pkg/pop/poptoken.go
rharpavat: Moved `PoPAuthenticationScheme` code to `authnscheme.go`

weinong:
please add UT to pkg/converter/convert.go

pkg/converter/convert.go
```diff
@@ -282,6 +309,12 @@ func Convert(o Options, pathOptions *clientcmd.PathOptions) error {
 				exec.Args = append(exec.Args, argEnvironment, argEnvironmentVal)
 			}
 
+			// poP token flags are optional but must be provided together
```
weinong, line 312: Pop or pop?

pkg/token/serviceprincipaltoken.go
```diff
@@ -69,91 +73,171 @@ func (p *servicePrincipalToken) Token() (adal.Token, error) {
...
+func (p *servicePrincipalToken) getTokenWithClientSecret(context context.Context, scopes []string, options *azcore.ClientOptions) (string, int64, error) {
+	clientOptions := &azidentity.ClientSecretCredentialOptions{
+		ClientOptions: azcore.ClientOptions{
+			Cloud: p.cloud,
+		},
+	}
+	if options != nil {
+		clientOptions.ClientOptions = *options
+	}
+	if p.popClaims != nil && len(p.popClaims) > 0 {
```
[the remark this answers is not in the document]
rharpavat: Nice catch, I didn't know this was possible in go - simplified all occurrences!

pkg/pop/poptokenutils.go
rharpavat: Added descriptions for all new funcs

pkg/pop/poptoken.go
rharpavat: Added structs to handle this

pkg/converter/convert.go
rharpavat: Updated the casing here

pkg/pop/authnscheme.go
```diff
@@ -0,0 +1,148 @@
...
+// type representing a PoP access token
+type PoPAccessToken struct {
+	Header    header
+	Payload   payload
+	Signature signature
+}
+
+// given a header, payload, and PoP key, creates the signature for the token and returns
+// a PoPAccessToken object representing the signed token
+func CreatePoPAccessToken(h header, p payload, popKey PoPKey) (*PoPAccessToken, error) {
```
julienstroheker, line 74: Do we need this to be exported ?

pkg/pop/authnscheme.go
```diff
@@ -0,0 +1,148 @@
...
+	sig []byte
+}
+
+// returns a base-64 encoded representation of a signature object
+func (s *signature) ToBase64() string {
+	return base64.RawURLEncoding.EncodeToString(s.sig)
+}
+
+// type representing a PoP access token
+type PoPAccessToken struct {
```
julienstroheker, line 66: Do we need this to be exported ?

pkg/pop/poptoken.go
```diff
@@ -0,0 +1,149 @@
...
+type swKey struct {
+	key    *rsa.PrivateKey
+	keyID  string
+	jwk    string
+	jwkTP  string
+	reqCnf string
+}
+
+// returns the algorithm used to encrypt/sign the swKey
+func (swk *swKey) Alg() string {
```
julienstroheker, line 45: There is couple function that can be UTs here

julienstroheker: Mainly to get a good CC

pkg/token/interactive.go
```diff
@@ -51,32 +54,55 @@ func (p *InteractiveToken) Token() (adal.Token, error) {
...
-		return emptyToken, errors.New("did not receive a token")
+	var token string
+	var expirationTimeUnix int64
+	var err error
+
+	if len(p.popClaims) > 0 {
+		// If PoP token support is enabled and the correct u-claim is provided, use the MSAL
+		// token provider to acquire a new token
+		token, expirationTimeUnix, err = pop.AcquirePoPTokenInteractive(
+			context.Background(),
```
julienstroheker, line 67: Usually this is a good practice to start a context as soon as we can in a code and make it flow (usefull for tracing for example). 

A bit out of scope of that PR, but we should at least declare a `ctx := context.Background()` at the top for the function and re-use it.

pkg/token/interactive.go
```diff
@@ -51,32 +54,55 @@ func (p *InteractiveToken) Token() (adal.Token, error) {
...
+			ClientOptions: clientOpts,
+			TenantID:      p.tenantID,
+			ClientID:      p.clientID,
+		})
+		if err != nil {
+			return emptyToken, fmt.Errorf("unable to create credential. Received: %w", err)
+		}
+
+		// Use the token provider to get a new token
+		interactiveToken, err := cred.GetToken(context.Background(), policy.TokenRequestOptions{Scopes: scopes})
```
julienstroheker, line 87: here you can re-use the `ctx` declared at the top (mentioned on my other comment)

pkg/token/options.go
```diff
@@ -118,6 +120,8 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.BoolVar(&o.IsLegacy, "legacy", o.IsLegacy, "set to true to get token with 'spn:' prefix in audience claim")
 	fs.BoolVar(&o.UseAzureRMTerraformEnv, "use-azurerm-env-vars", o.UseAzureRMTerraformEnv,
 		"Use environment variable names of Terraform Azure Provider (ARM_CLIENT_ID, ARM_CLIENT_SECRET, ARM_CLIENT_CERTIFICATE_PATH, ARM_CLIENT_CERTIFICATE_PASSWORD, ARM_TENANT_ID)")
+	fs.BoolVar(&o.IsPoPTokenEnabled, "pop-enabled", o.IsPoPTokenEnabled, "set to true to use a PoP token for authentication or false to use a regular bearer token")
```
julienstroheker, line 123: I assume this is defaulted to false when not been set right ?

pkg/token/provider.go
```diff
@@ -24,13 +24,17 @@ func newTokenProvider(o *Options) (TokenProvider, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to get cloud.Configuration. err: %s", err)
 	}
+	popClaimsMap, err := parsePopClaims(o.PoPTokenClaims)
+	if o.IsPoPTokenEnabled && err != nil {
+		return nil, fmt.Errorf("failed to parse pop token claims. err: %w", err)
```
julienstroheker, line 29: the error msg might be a duplicate based on what `parsePopClaims` function will returns.

`failed to parse pop token claims. err: error parsing PoP token claims. Ensure the claims are formatted as `key=value` with no extra whitespace`

pkg/token/serviceprincipaltoken.go
```diff
@@ -69,91 +73,177 @@ func (p *servicePrincipalToken) Token() (adal.Token, error) {
...
-			p.clientSecret,
-			clientOptions,
-		)
-		if err != nil {
-			return emptyToken, fmt.Errorf("unable to create credential. Received: %w", err)
-		}
-
-		// Use the token provider to get a new token
-		spnAccessToken, err = cred.GetToken(context.Background(), policy.TokenRequestOptions{Scopes: []string{p.resourceID + "/.default"}})
+		accessToken, expirationTimeUnix, err = p.getTokenWithClientSecret(context.Background(), scopes, options)
```
julienstroheker, line 83: same comment about the `context.Background()`

pkg/token/serviceprincipaltoken.go
```diff
@@ -69,91 +73,177 @@ func (p *servicePrincipalToken) Token() (adal.Token, error) {
...
+		p.tenantID,
+		options,
+	)
+	if err != nil {
+		return "", -1, fmt.Errorf("failed to create service principal PoP token using certificate: %w", err)
+	}
+
+	return accessToken, expiresOn, nil
+}
+
```
julienstroheker, line 246: @weinong should we split this into 2 files `serviceprincipaltokensecret.go` and `serviceprincipaltokencert.go`, this is starting to be a lot of lines ?

julienstroheker (changes requested):
PTAL at my comments/questions

pkg/token/options.go
rharpavat: Yeah it is only set to true if the flag was provided

pkg/token/interactive.go
rharpavat: Good point, updated to use a single ctx for all the calls in this func

pkg/token/interactive.go
rharpavat: Updated to reuse the ctx

pkg/token/provider.go
rharpavat: Updated the error messages in `parsePopClaims` and removed the additional error wrapping here since those are already descriptive

pkg/token/serviceprincipaltoken.go
rharpavat: Updated to reuse ctx

pkg/pop/authnscheme.go
rharpavat: Changed this to be unexported since it's not used outside this file

pkg/pop/authnscheme.go
rharpavat: Changed this to be unexported since it's not used outside this file

pkg/pop/msal.go
```diff
@@ -0,0 +1,89 @@
...
+	"fmt"
+	"net/http"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/confidential"
+	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/public"
+)
+
+// acquires a PoP token using MSAL's interactive login flow. Requires user to authenticate via browser
+func AcquirePoPTokenInteractive(
```
weinong, line 17: why does it not have `options *azcore.ClientOptions` like below?

pkg/pop/msal.go
```diff
@@ -0,0 +1,89 @@
+// Copyright (c) Microsoft Corporation.
```
weinong, line 1: any unit test?

weinong:
For exported functions, i'd recommend adhering to golang coding convention to start comment with function name.

pkg/pop/poptoken.go
weinong: +1

pkg/pop/poptoken.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	key, err := generateSwKey()
+	if err != nil {
+		log.Panicf("unable to generate popkey. err: %v", err)
+	}
+	pswKey = key
+
+	// rotate key every 8 hours
+	ticker := time.NewTicker(8 * time.Hour)
```
weinong, line 134: kubelogin is a cli, right? how does it live for more than 8 hours?

pkg/pop/poptoken.go
```diff
@@ -0,0 +1,149 @@
...
+
+	swk.init(key)
+	return swk, nil
+}
+
+var pswKey *swKey
+var pwsKeyMutex sync.Mutex
+
+// generates a new PoP key that rotates every 8 hours and returns it
+func GetSwPoPKey() *swKey {
```
weinong, line 120: can it return error so that it doesn't panic?

pkg/token/serviceprincipaltoken.go
weinong: sounds like a good idea!

pkg/token/testdata/ServicePrincipalPoPTokenFromBadPoPClaimsVCR.yaml
```diff
@@ -0,0 +1,174 @@
...
+        transfer_encoding: []
+        trailer: {}
+        host: login.microsoftonline.com
+        remote_addr: ""
+        request_uri: ""
+        body: ""
+        form: {}
+        headers:
+            User-Agent:
+                - microsoft-authentication-library-for-go/v1.1.1-0.20230811173331-99e839c4c840 (go1.19.6; linux)
```
weinong, line 19: can you add kubelogin identifier to the user agent?

pkg/pop/poptoken.go
rharpavat: when running any `kubectl` commands, if user is not authenticated, it will call `kubelogin` under the hood to get a token and then cache the token somewhere + reuse until auth fails again. I guess this also applies if user runs `kubelogin get-token` manually and sets that token in the kubeconfig

pkg/token/serviceprincipaltoken.go
rharpavat: Split this into 3 files and added additional UTs to increase coverage

pkg/pop/poptoken.go
rharpavat: Updated this func to return error in case of failure, plus added error handling for the returned errors

pkg/pop/msal.go
rharpavat: Added the client options here

pkg/token/testdata/ServicePrincipalPoPTokenFromBadPoPClaimsVCR.yaml
julienstroheker: @weinong do you mean to makes sure "live" call to msal is setting a kubelogin user-agent or just overwriting the vcr cassette ?

Is this is about real scenario; msal does not expose a helper function to do that. Only way I can think about is to pass a custom httclient with a round tripper. This will requires more code.

weinong: yes, live calls. huh... azidentity can do that. so i suppose msal can as well? if not possible, then forget about it

pkg/pop/poptoken.go
weinong: as soon as kubelogin returns the token, the process exited. how does it cache in memory?

pkg/token/testdata/ServicePrincipalPoPTokenFromBadPoPClaimsVCR.yaml
julienstroheker: It's is doable but not as easy as azidentity. IMO this can be a separated PR if really needed.

pkg/pop/poptoken.go
rharpavat: Ah the token is cached on disk under the folder `~/.kube/cache/`. Though in our scenario, the token caching doesn't really matter since Guard is configured to reject the token after 15 mins regardless of the token's expiry.

pkg/pop/poptoken.go
rharpavat: Added UTs for this file!

pkg/pop/msal.go
rharpavat: Added some VCR tests for the confidential flow

pkg/token/testdata/ServicePrincipalPoPTokenFromBadPoPClaimsVCR.yaml
rharpavat: Will resolve this comment as there are no action items

pkg/pop/authnscheme.go
```diff
@@ -0,0 +1,152 @@
...
+
+// ToBase64 returns a base-64 encoded representation of a PoP access token
+func (p *popAccessToken) ToBase64() string {
+	return fmt.Sprintf("%s.%s.%s", p.Header.ToBase64(), p.Payload.ToBase64(), p.Signature.ToBase64())
+}
+
+// PoPAuthenticationScheme is a PoP token implementation of the MSAL AuthenticationScheme interface
+type PoPAuthenticationScheme struct {
+	// host is the u claim we will add on the pop token
+	Host   string
```
julienstroheker, line 98: You should add a comment saying that this interface will only take care of the `u` claims been passed. If anybody is trying to pass more claims this will not be take in consideration

pkg/pop/poptoken.go
julienstroheker: @weinong point is fair, this code is not needed on kubelogin since as cli it is a one shot operation then memory will be released and this routine will never been executed. 

This code is more when running this library on a server side.

pkg/pop/poptoken.go
```diff
@@ -0,0 +1,178 @@
+// Copyright (c) Microsoft Corporation.
```
julienstroheker, line 1: Seeing some file with this header, other without. Should we remove this @weinong ?

pkg/testutils/testutils.go
```diff
@@ -0,0 +1,23 @@
...
+	ClientID       = "AZURE_CLIENT_ID"
+	ClientSecret   = "AAD_SERVICE_PRINCIPAL_CLIENT_SECRET"
+	ClientCert     = "AZURE_CLIENT_CER"
+	ClientCertPass = "AZURE_CLIENT_CERTIFICATE_PASSWORD"
+	ResourceID     = "AZURE_RESOURCE_ID"
+	TenantID       = "AZURE_TENANT_ID"
+	BadSecret      = "Bad_Secret"
+)
+
+func ErrorContains(out error, want string) bool {
```
julienstroheker, line 15: this can be UT

pkg/pop/poptoken.go
rharpavat: Removed this code since it's not actually being used in the CLI

pkg/pop/authnscheme.go
```diff
@@ -0,0 +1,152 @@
...
+	return token, nil
+}
+
+// ToBase64 returns a base-64 encoded representation of a PoP access token
+func (p *popAccessToken) ToBase64() string {
+	return fmt.Sprintf("%s.%s.%s", p.Header.ToBase64(), p.Payload.ToBase64(), p.Signature.ToBase64())
+}
+
+// PoPAuthenticationScheme is a PoP token implementation of the MSAL AuthenticationScheme interface
+type PoPAuthenticationScheme struct {
```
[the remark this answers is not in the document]
rharpavat: Reordered the funcs

pkg/pop/authnscheme.go
rharpavat: Added comment + disclaimer

pkg/testutils/testutils.go
rharpavat: Added UTs for this func

pkg/pop/poptoken.go
```diff
@@ -0,0 +1,159 @@
...
+
+// generateSwKey generates a new swkey and initializes it with required fields before returning it
+func generateSwKey(key *rsa.PrivateKey) (*swKey, error) {
+	swk := &swKey{}
+	swk.init(key)
+	return swk, nil
+}
+
+var pswKey *swKey
+var pwsKeyMutex sync.Mutex
```
julienstroheker, line 96: you don't need this anymore

pkg/pop/poptoken.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// generateSwKey generates a new swkey and initializes it with required fields before returning it
+func generateSwKey(key *rsa.PrivateKey) (*swKey, error) {
+	swk := &swKey{}
+	swk.init(key)
+	return swk, nil
+}
+
+var pswKey *swKey
```
julienstroheker, line 95: you don't need this anymore

pkg/pop/poptoken.go
```diff
@@ -0,0 +1,159 @@
...
+func GetSwPoPKey() (*swKey, error) {
+	key, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		return nil, fmt.Errorf("error generating RSA private key: %w", err)
+	}
+	return GetSwPoPKeyWithRSAKey(key)
+}
+
+func GetSwPoPKeyWithRSAKey(rsaKey *rsa.PrivateKey) (*swKey, error) {
+	pwsKeyMutex.Lock()
```
julienstroheker, line 108: the lock `pwsKeyMutex` and the global key `pswKey` can be removed since you are not using this anymore.

this funciton can just be 

```
	key, err := generateSwKey(rsaKey)
	if err != nil {
		return nil, fmt.Errorf("unable to generate popkey. err: %w", err)
	}
	return key, nil
```

pkg/pop/authnscheme.go
```diff
@@ -0,0 +1,155 @@
...
+	"strings"
+	"time"
+
+	"github.com/google/uuid"
+)
+
+// type of a PoP token, as opposed to "JWT" for a regular bearer token
+const popTokenType = "pop"
+
+// PoPAuthenticationScheme is a PoP token implementation of the MSAL AuthenticationScheme interface.
```
julienstroheker, line 20: you should mention `PoPAuthenticationScheme is a PoP token implementation of the MSAL AuthenticationScheme interface used by the Azure Arc Platform team`

julienstroheker (changes requested):
PTAL

pkg/pop/authnscheme.go
rharpavat: Updated the text to be more specific

pkg/pop/poptoken.go
rharpavat: Cleaned up the function

pkg/pop/poptoken.go
rharpavat: removed

pkg/pop/poptoken.go
rharpavat: removed

julienstroheker (approved):

pkg/pop/poptoken.go
weinong: sure. let's remove that for consistency

weinong (approved):
LGTM

pkg/pop/poptoken.go
rharpavat: Removed the header from code files

PR #338, opened 2023-09-13
use the adal library for spn when --legacy is specified

### LINKED ISSUE #331: legacy flag does not work from version v0.0.30 when the login method is spn
How to reproduce it

kubelogin get-token --tenant-id <<tenant-id>> --server-id <<server-id>> --client-id <<client-id>> --client-secret <<client-secret>> --login spn --legacy

Expected outcome:

token with audience field has "spn:" prefix

It works fine  on version v0.0.29 and lower

weinong:
it will fix #331

### DIFF
--- a/pkg/token/serviceprincipaltoken_legacy.go
+++ b/pkg/token/serviceprincipaltoken_legacy.go
@@ -53,9 +53,6 @@ func newLegacyServicePrincipalToken(oAuthConfig adal.OAuthConfig, clientID, clie
 
 func (p *legacyServicePrincipalToken) Token() (adal.Token, error) {
 	emptyToken := adal.Token{}
-	callback := func(t adal.Token) error {
-		return nil
-	}
 
 	var (
 		spt *adal.ServicePrincipalToken
@@ -67,8 +64,7 @@ func (p *legacyServicePrincipalToken) Token() (adal.Token, error) {
 			p.oAuthConfig,
 			p.clientID,
 			p.clientSecret,
-			p.resourceID,
-			callback)
+			p.resourceID)
 		if err != nil {
 			return emptyToken, fmt.Errorf("failed to create service principal token using secret: %s", err)
 		}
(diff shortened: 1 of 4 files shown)


pkg/token/serviceprincipaltoken_legacy.go
```diff
@@ -0,0 +1,118 @@
...
+			return emptyToken, fmt.Errorf("failed to decode pkcs12 certificate while creating spt: %w", err)
+		}
+
+		spt, err = adal.NewServicePrincipalTokenFromCertificate(
+			p.oAuthConfig,
+			p.clientID,
+			cert,
+			rsaPrivateKey,
+			p.resourceID,
+			callback)
```
bcho, line 93: https://pkg.go.dev/github.com/Azure/go-autorest/autorest/adal#NewServicePrincipalTokenFromCertificate

the callback here is optional, since it's using an empty implementation here, I think we can drop it?

pkg/token/serviceprincipaltoken_legacy.go
```diff
@@ -0,0 +1,118 @@
...
+		err error
+	)
+
+	if p.clientSecret != "" {
+		spt, err = adal.NewServicePrincipalToken(
+			p.oAuthConfig,
+			p.clientID,
+			p.clientSecret,
+			p.resourceID,
+			callback)
```
bcho, line 71: similar to comment in below

pkg/token/serviceprincipaltoken_legacy.go
```diff
@@ -0,0 +1,118 @@
...
+			cert,
+			rsaPrivateKey,
+			p.resourceID,
+			callback)
+		if err != nil {
+			return emptyToken, fmt.Errorf("failed to create service principal token using cert: %s", err)
+		}
+	}
+
+	err = spt.Refresh()
```
bcho, line 99: How about `EnsureFreshWithContext` so it can handle different kind of token expiry cases?

weinong: this token provider doesn't cache, so it always gets fresh token

weinong: but.. why not!

bcho (approved):
LGTM

PR #342, opened 2023-09-26
Update install.md

torreymicrosoft:
Per email thread with weinong, proposing a PR to provide additional pathway to install kubelogin on Linux via https://learn.microsoft.com/en-us/cli/azure/aks?view=azure-cli-latest#az-aks-install-cli(). Providing suggested additional pathway, since asdf seems to encounter installation difficulty on Linux and with AKS. Creating PR for this.

This is also briefly discussed at the bottom of the issue #173.


weinong (approved):
LGTM. thanks!

Issue #353, opened 2023-10-24
Support OIDC request url & token


stevehipwell:
I'd like `kubelogin` to support the `ACTIONS_ID_TOKEN_REQUEST_URL` & `ACTIONS_ID_TOKEN_REQUEST_TOKEN` environmental variables to handle OIDC authentication (in the same way that the AzureRM TF provider works).

Currently I create the token just before running Terraform but if there is a long process before `kubelogin` is called (such as creating an AKS cluster) the OIDC token has expired and Terraform fails.
```[tasklist]
### Tasks
```

weinong:
i like the idea!

stevehipwell:
This is starting to cause us a significant number of failures. I'm going to look at wrapping `kubelogin` with a script but having this as native functionality would be desirable.

PR #366, opened 2023-12-01
added github token support

weinong:
added github token support via ACTIONS_ID_TOKEN_REQUEST_URL and ACTIONS_ID_TOKEN_REQUEST_TOKEN in workload identity login

it addresses #353

### DIFF
--- a/pkg/token/federatedIdentity.go
+++ b/pkg/token/federatedIdentity.go
@@ -38,8 +38,9 @@ func newWorkloadIdentityToken(clientID, federatedTokenFile, authorityHost, serve
 	if tenantID == "" {
 		return nil, errors.New("tenantID cannot be empty")
 	}
-	if federatedTokenFile == "" && (os.Getenv(actionsIDTokenRequestToken) == "" || os.Getenv(actionsIDTokenRequestURL) == "") {
-		return nil, fmt.Errorf("either %s and %s environment variables have to be set or federated token file has to be provided", actionsIDTokenRequestToken, actionsIDTokenRequestURL)
+	hasActionsIDToken := os.Getenv(actionsIDTokenRequestToken) != "" && os.Getenv(actionsIDTokenRequestURL) != ""
+	if federatedTokenFile == "" && !hasActionsIDToken {
+		return nil, errors.New("either ACTIONS_ID_TOKEN_REQUEST_TOKEN and ACTIONS_ID_TOKEN_REQUEST_URL environment variables have to be set or federated token file has to be provided")
 	}
 	if authorityHost == "" {
 		return nil, errors.New("authorityHost cannot be empty")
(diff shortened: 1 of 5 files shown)


weinong:
tested in [this github actions](https://github.com/weinong/azure-federated-identity-samples/actions/runs/7056224233/job/19208097992).

pkg/token/federatedIdentity.go
```diff
@@ -27,8 +38,8 @@ func newWorkloadIdentityToken(clientID, federatedTokenFile, authorityHost, serve
 	if tenantID == "" {
 		return nil, errors.New("tenantID cannot be empty")
 	}
-	if federatedTokenFile == "" {
-		return nil, errors.New("federatedTokenFile cannot be empty")
+	if federatedTokenFile == "" && (os.Getenv(actionsIDTokenRequestToken) == "" || os.Getenv(actionsIDTokenRequestURL) == "") {
+		return nil, fmt.Errorf("either %s and %s environment variables have to be set or federated token file has to be provided", actionsIDTokenRequestToken, actionsIDTokenRequestURL)
```
bcho, line 42: nit: this error can be extracted as a global value?

pkg/token/federatedIdentity.go
```diff
@@ -114,3 +111,54 @@ func readJWTFromFS(tokenFilePath string) (string, error) {
 	}
 	return string(token), nil
 }
+
+func getGitHubToken() (string, error) {
```
bcho, line 115: let's pass the context into this function since it's doing an external http call

pkg/token/federatedIdentity.go
```diff
@@ -114,3 +111,54 @@ func readJWTFromFS(tokenFilePath string) (string, error) {
...
+	client := http.Client{}
+	resp, err := client.Do(req)
+	if err != nil {
+		return "", err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		body, _ := io.ReadAll(resp.Body)
+		return "", fmt.Errorf("github actions ID token request failed with status code: %d, response body: %s", resp.StatusCode, body)
```
bcho, line 151: the body here might be nil when io.ReadAll failed. Passing a nil value to `%s` is unexpected. And the type of `body` is `[]byte`, I assume you mean to `string(body)` instead? Maybe we can do this:

```
body, readBodyErr := io.ReadAll(resp.Body)
var bodyString string
if readBodyErr == nil {
  bodyString = string(body)
} else {
  bodyString = readBodyErr.Error()
}
```

pkg/token/federatedIdentity.go
weinong: i can change it to errors.New(), but what's the value of global var?

bcho (approved):
LGTM

pkg/token/federatedIdentity.go
weinong: i didn't opt to refactor this as there are many other errors.New() already

Issue #373, opened 2023-12-12
Proposal: kubelogin library usage


bcho:
This proposal provides an overview of enabling kubelogin to be used as a library.

## Background

There are a few asks on importing kubelogin as library from both internal and external users. The latest one is https://github.com/Azure/kubelogin/pull/371. Here is a list of known library imports ([ref](https://pkg.go.dev/github.com/Azure/kubelogin@v0.0.34/pkg/token?tab=importedby)):

- [argocd-k8s-auth](https://github.com/argoproj/argo-cd/blob/6eba5be864b7e031871ed7698f5233336dfe75c7/cmd/argocd-k8s-auth/commands/azure.go#L6)

  * importing the external credentials plugin
- [crossplane provider-helm](https://github.com/crossplane-contrib/provider-helm/blob/6ba133b74d57e59ccd02f3ae2fdf373309a6534e/pkg/clients/azure/azure.go#L8)
  * importing the get token function (via gomod replace)

The original design of kubelogin is to be consumed as a command line tool via the [external credentials provider protocol](https://github.com/kubernetes/enhancements/blob/master/keps/sig-auth/541-external-credential-providers/README.md). However, for 3rd party Kubernetes DevOps services, it's a common task to embed the same authentication flow into the product without the need of downloading kubelogin binary. In this case, we should enable the 3rd party service to import kubelogin as a library.

There are a few blockers for enabling this usage:
- `getToken` call is not being exposed
- `TokenProvider` is returning legacy `adal.Token` object
- `token.Options` exposes CLI usage settings

### `getToken` call is not being exposed

The most needed feature is to generate a token using user environment input. We can do it via `getToken` call. However, this function is private to 3rd party library usage. We should expose this function along with the options structure.

### `TokenProvider` is returning legacy `adal.Token` object

Even though kubelogin is making use of `azidentity` internally, for local caching compatibility, the `getToken` function and `TokenProvider` interface are still returning the legacy `adal.Token` object. To avoid introducing legacy dependencies to 3rd party services, we should not expose this legacy interface anymore. Ideally, a string only representation should be used instead of pinning on a specific Azure SDK version as his approach can decouple the dependencies.

### `token.Options` exposes CLI usage settings

[...]

PR #376, opened 2023-12-18
[library usage] Move modules under `pkg` to `pkg/internal`

bcho:

In this pull request we moved most of the sub packages under `pkg` to `pkg/internal` except the `pkg/cmd`. The `pkg/cmd` is needed to be exported from `pkg` level as Go's internal package rule limit the internal package visibility level to `pkg` only. We also hid sub-commands from `pkg/cmd` to avoid unexpected references.

Now, the root level main function imports only `pkg/cmd` and it will be served as the sole entry point. This is slightly off from the design doc, but I believe the usage experience would be better as we allow user to install kubelogin via this:

```
$ go install github.com/Azure/kubelogin@{version}
```


weinong (approved):
LGTM

PR #379, opened 2023-12-18
[library usage] Refine internal token types

bcho:

In this pull request, we

- refactored the `TokenProvider` interface to accept `context.Context`. This allows library user to propagate the context properly. All methods using `TokenProvider` are being updated as well.
- exported `NewTokenProvider` method later library usage.
- bump gomock to uber go mock version as the original one has been deprecated.


weinong (approved):
LGTM

PR #380, opened 2023-12-19
[library usage] Implement library token provider

bcho:

In this pull request, we implemented the library token provider instance and options settings. To allow reusing the environment variables, we introduced a new package `internal/env` for holding all environment variables.


pkg/token/options.go
```diff
@@ -0,0 +1,47 @@
+package token
+
+import "github.com/Azure/kubelogin/pkg/internal/token"
+
+// list of supported login methods for library consumers
+
+const (
+	ServicePrincipalLogin = token.ServicePrincipalLogin
+	ROPCLogin             = token.ROPCLogin
```
weinong, line 9: i think we should drop ropc unless someone requested it

pkg/token/types.go
```diff
@@ -0,0 +1,16 @@
...
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+)
+
+// AccessToken represents an Azure service bearer access token with expiry information.
+type AccessToken = azcore.AccessToken
+
+// TokenProvider provides access to tokens.
+type TokenProvider interface {
+	// GetAccessToken returns an access token from giving settings.
```
weinong, line 14:
```suggestion
	// GetAccessToken returns an access token from given settings.
```

pkg/token/options.go
bcho: done, dropped the `ROPCLogin` method

weinong (approved):
Left a non-blocking comment.

PR #386, opened 2023-12-28
[library usage] fix: downgrade required go version to 1.20

bcho:
When using kubelogin as library, the go version specified in go.mod becomes the minimum required version to the library consumer. To allow further reuse, this pr downgrades the go version to 1.20 to align with Go's official support policy, which supports the last two major versions.


weinong (approved):
LGTM

PR #398, opened 2024-01-19
Adds Azure Developer CLI (azd) as a new login method

wbreza:
Adds support for Azure Developer CLI (azd) to be used as a login method.  

When users are deploying AKS based application with `azd`, the kubeconfig will automatically be converted to use `azd` authentication when working with clusters with RBAC enabled and local user accounts disabled.

Also updates `azidentity` package to `v1.5.1` to leverage new `AzureDeveloperCLICredential`

## What is Azure Developer CLI

https://aka.ms/azd

Azure Developer CLI (azd) is an open-source tool that accelerates the time it takes for you to get your application from local development environment to Azure. azd provides best practice, developer-friendly commands that map to key stages in your workflow, whether you're working in the terminal, your editor or integrated development environment (IDE), or CI/CD (continuous integration/continuous deployment).


pkg/internal/token/azuredevelopercli.go
```diff
@@ -0,0 +1,81 @@
...
+		timeout:    timeout,
+	}, nil
+}
+
+// Token fetches an azcore.AccessToken from the Azure CLI SDK and converts it to an adal.Token for use with kubelogin.
+func (p *AzureDeveloperCLIToken) Token(ctx context.Context) (adal.Token, error) {
+	emptyToken := adal.Token{}
+
+	// Request a new Azure CLI token provider
+	cred, err := azidentity.NewAzureDeveloperCLICredential(&azidentity.AzureDeveloperCLICredentialOptions{
```
weinong, line 45: you probably copied the pattern from azurecli implementation. the credential should be created in newFunction like https://github.com/Azure/kubelogin/blob/main/pkg/internal/token/federatedIdentity.go#L59. it didn't have any material impact until recently we enabled library usage.

docs/book/src/cli/convert-kubeconfig.md
weinong, line 1: can you also add a doc under https://github.com/Azure/kubelogin/tree/main/docs/book/src/concepts/login-modes

pkg/internal/token/azuredevelopercli.go
wbreza: The internal credential is being created via the updated `azidentity` package and `NewAzureDeveloperCLICredential` new function.  Let me know what you are expecting to be different here.

weinong: since in library use client may keep the token provider longer and call `Token()` as many times as needed, it's better to save the `cred` in `AzureDeveloperCLIToken` struct

weinong: your PR is not adding this login to library use, which is fine. but i would still like this credential  to be created in new function. :)

pkg/internal/converter/convert.go
```diff
@@ -253,6 +253,11 @@ func Convert(o Options, pathOptions *clientcmd.PathOptions) error {
 		}
 
 		switch o.TokenOptions.LoginMethod {
+		case token.AzureDeveloperCLILogin:
```
weinong, line 256: please add a unit test

wbreza: @weinong Added a couple valid use cases for `azd`.  I did not add all the legazy or override use cases since there should not be any special support required for these.

If you would prefer me to add them just let me know.

weinong (approved):
LGTM

PR #447, opened 2024-04-18
Move PoP outside of Internal Folder

JorgeDaboub:
Within the Arc Connectivity team, we make use of CSP (client-side proxy). 

A sender component used to send messages for our scenarios. The code we use for our Pop scenario is the same one as PoP. 

We are collaborating with @julienstroheker to maintain only a single instance of this code. And thus would require the ability to access this module from our CSP component.


bcho:
Hi, can you tell us a bit more why this change is needed? Also, we suggest starting with an issue for this kind of discussions.

pkg/internal/token/interactive.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	"github.com/Azure/go-autorest/autorest/adal"
-	"github.com/Azure/kubelogin/pkg/internal/pop"
+	"github.com/Azure/kubelogin/pkg/pop"
```
bcho, line 15: we cannot let internal package to import a public package as this is might be troublesome when there is cyclic imports and api changes.

JorgeDaboub:
> Hi, can you tell us a bit more why this change is needed? Also, we suggest starting with an issue for this kind of discussions.

Hi Bcho, added context. Accidentality opened the PR before adding it.

bcho:
@JorgeDaboub do you have the exact methods that needed? I would suggest we do the similar thing like this:

https://github.com/Azure/kubelogin/blob/6ad3af502d133b1795936b77fafc0816738cc17f/pkg/token/options.go#L8-L10

to create a package under `pkg/` like `pkg/pop`, then put the required methods in there, but keep the implementation inside of the internal package. This can minimize the exposure while providing the usages as library.

PR #476, opened 2024-06-12
feat: declare go version directive with patch version

bcho:
We need this change for making dependabot to understand the toolchain requirements.

Ref: https://go.dev/doc/toolchain

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,7 @@
 module github.com/Azure/kubelogin
 
 // NOTE: kubelogin follows the same support policy as Go, which supports the last two major versions.
-go 1.21.0
+go 1.21.9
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -1,7 +1,7 @@
 module github.com/Azure/kubelogin
 
 // NOTE: kubelogin follows the same support policy as Go, which supports the last two major versions.
-go 1.21
+go 1.21.0
```
weinong, line 4: should it be one of the most recent 1.21 instead? its currently built using 1.21.9

bcho: updated

weinong (approved):
LGTM

Issue #500, opened 2024-08-02
azure kubelogin doesn't respect KUBECACHEDIR


jjournet:
Hello,

[kubectl supports the KUBECACHEDIR](https://github.com/kubernetes/kubernetes/pull/109479) environment variable to change the default location of `~/.kube/cache`

The kubelogin implementaion from int128 fixed its implementation and [respects now the KUBECACHEDIR env var](https://github.com/int128/kubelogin/pull/975).

Would it be possible to respect the KUBECACHEDIR environment variable in azure kubelogin ?

If a use case helps justify the need: I am using [environment module](https://modules.readthedocs.io/en/latest/) to load different configuration for az cli and kubectl, as we are using different az identities for multiple clusters.
With KUBECACHEDIR not being supported, I can only use one az identity at a time for kubectl, as `~/.kube/cache` is shared, whereas az and kubectl config are separated.

Thank you
JP

PR #501, opened 2024-08-02
support of environment variable KUBECACHEDIR #500

jjournet:
Fix to support environment variable KUBECACHEDIR, already supported by kubectl, to change the location of folder `~/.kube/cache`.
This is necessary to support having different environments using different Azure identities.

### DIFF
--- a/pkg/internal/token/options.go
+++ b/pkg/internal/token/options.go
@@ -63,13 +63,16 @@ func GetSupportedLogins() string {
 }
 
 func NewOptions() Options {
-	if env_TokenCacheDir := os.Getenv("KUBECACHEDIR"); env_TokenCacheDir != "" {
-		DefaultTokenCacheDir = env_TokenCacheDir
-	}
+	env_TokenCacheDir := os.Getenv("KUBECACHEDIR")
 	return Options{
-		LoginMethod:   DeviceCodeLogin,
-		Environment:   defaultEnvironmentName,
-		TokenCacheDir: DefaultTokenCacheDir,
+		LoginMethod: DeviceCodeLogin,
+		Environment: defaultEnvironmentName,
+		TokenCacheDir: func() string {
+			if env_TokenCacheDir != "" {
+				return env_TokenCacheDir
+			}
+			return DefaultTokenCacheDir
+		}(),
 	}
 }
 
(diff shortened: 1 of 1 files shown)


jjournet:
fix #500

pkg/internal/token/options.go
```diff
@@ -63,10 +63,16 @@ func GetSupportedLogins() string {
...
 
 func NewOptions() Options {
+	env_TokenCacheDir := os.Getenv("KUBECACHEDIR")
 	return Options{
-		LoginMethod:   DeviceCodeLogin,
-		Environment:   defaultEnvironmentName,
-		TokenCacheDir: DefaultTokenCacheDir,
+		LoginMethod: DeviceCodeLogin,
+		Environment: defaultEnvironmentName,
+		TokenCacheDir: func() string {
```
weinong, line 70: you can just check if `env_TokenCacheDir` is empty or not right after L66 and set it accordingly.

jjournet: does the change follow what you had in mind @weinong ?

if yes, I'll resolver the conversation.
thank you for your advice

pkg/internal/token/options.go
```diff
@@ -63,6 +63,9 @@ func GetSupportedLogins() string {
 }
 
 func NewOptions() Options {
+	if env_TokenCacheDir := os.Getenv("KUBECACHEDIR"); env_TokenCacheDir != "" {
```
weinong, line 66: let's not change the global variable. how about this?

```go
func NewOptions() Options {
	tokenCacheDir := DefaultTokenCacheDir
	if envTokenCacheDir := os.Getenv("KUBECACHEDIR"); envTokenCacheDir != "" {
		tokenCacheDir = envTokenCacheDir
	}
	return Options{
		LoginMethod:   DeviceCodeLogin,
		Environment:   defaultEnvironmentName,
		TokenCacheDir: tokenCacheDir,
	}
}
```

jjournet: @weinong that's how I implemented it my first commit.

The difference is the use of the inline function to avoid adding a new variable, and I find it more elegant.

I pushed a revert

weinong (approved):
LGTM

jjournet:
@weinong thank you for the approval

lint didn't like my variable, I changed it, waiting for your approval to run the actions again (sorry for the mistake)

PR #522, opened 2024-09-13
Use AZURE_CONFIG_DIR in kubelogin command example

tspearconquest:
When calling `kubelogin` with `AZURE_CONFIG_DIR` already set in a Jenkins job, just use that variable on the command line for consistency with the variable Jenkins is using, instead of doing `$WORKSPACE/.azure` which might not be where AZURE_CONFIG_DIR points to.

### DIFF
--- a/docs/book/src/topics/jenkins.md
+++ b/docs/book/src/topics/jenkins.md
@@ -12,7 +12,7 @@ Using kubelogin `convert-kubeconfig` subcommand with `--azure-config-dir`, the g
 stage('Download kubeconfig and convert') {
     steps {
         sh 'az aks get-credentials -g ${RESOURCE_GROUP} -n ${CLUSTER_NAME}'
-        sh 'kubelogin convert-kubeconfig -l azurecli --azure-config-dir ${WORKSPACE}/.azure'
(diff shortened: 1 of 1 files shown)


docs/book/src/topics/jenkins.md
```diff
@@ -12,7 +12,7 @@ Using kubelogin `convert-kubeconfig` subcommand with `--azure-config-dir`, the g
 stage('Download kubeconfig and convert') {
     steps {
         sh 'az aks get-credentials -g ${RESOURCE_GROUP} -n ${CLUSTER_NAME}'
-        sh 'kubelogin convert-kubeconfig -l azurecli --azure-config-dir ${WORKSPACE}/.azure'
+        sh 'kubelogin convert-kubeconfig -l azurecli --azure-config-dir ${AZURE_CONFIG_DIR}'
```
weinong, line 15: how about `${AZURE_CONFIG_DIR:-${WORKSPACE}/.azure}`?

tspearconquest: Hello, thanks for this suggestion and apologies for the delay. I've applied the suggested text.

weinong (approved):
LGTM

PR #542, opened 2024-10-31
Expose MSAL PoP for Consistent CSP Integration

JorgeDaboub:
This PR is needed to unify the custom PoP token implementation in the MSAL library for kubelogin and CSP (Client-Side Proxy), an Arc component used for line-of-sight and authentication with Arc resources.

@julienstroheker 

Changes in this PR:

- Exposed Methods:
  - AcquirePoPTokenConfidential
  - GetSwPoPKey
- Exposed Structs and Methods:
  - PoPAuthenticationScheme and its FormatAccessToken method.
  - SwKey and associated methods.
- Updated AcquirePoPTokenConfidential to accept popKey as a parameter, enabling reuse of the same key over time instead of generating a new one for each call, per our scenario's requirements.


JorgeDaboub:
This PR is needed to unify the custom PoP token implementation in the MSAL library for kubelogin and CSP (Client-Side Proxy), an Arc component used for line-of-sight and authentication with Arc resources.

pkg/internal/pop/poptoken.go
```diff
@@ -81,23 +88,57 @@ func (swk *swKey) init(key *rsa.PrivateKey) {
...
+	}
+
+	key, err := GetSwPoPKey()
+	if err != nil {
+		return nil, fmt.Errorf("error generating PoP key: %w", err)
+	}
+	pswKey = key
+
+	// rotates key every 8 hours
+	ticker := time.NewTicker(8 * time.Hour)
```
julienstroheker, line 114: In the context of kubelogin not sure if this make sense.
Kubelogin is a one shot binary invokation, you cannot maintain a go routine in the background for 8h.
For the kubelogin flow, we might want this to be unique all the time, if you need to change this behavior this will need to init/passed by the caller

pkg/pop/msal_confidential.go
```diff
@@ -0,0 +1,23 @@
...
+func AcquirePoPTokenConfidential(
+	context context.Context,
+	popClaims map[string]string,
+	scopes []string,
+	cred confidential.Credential,
+	authority,
+	clientID,
+	tenantID string,
+	options *azcore.ClientOptions,
+	popKey *SwKey,
```
julienstroheker, line 20: Based on my previous comment, this might need a to be function instead of a struct. This way on your side you can set your timer and other caller (like kubelogin) can decide what logic they want to do such as having an unique key all the time

pkg/pop/types.go
```diff
@@ -0,0 +1,25 @@
+package pop
+
+import (
+	"github.com/Azure/kubelogin/pkg/internal/pop"
+)
+
+type PoPAuthenticationScheme struct {
```
julienstroheker, line 7: add a comment on top of this that describe this is the msal implementation of AuthenticationScheme (You should add a link to msal repo interface)

pkg/pop/types.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+// This is the MSAL implementation of AuthenticationScheme.
+// For more details, see the MSAL repo interface:
+// https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/4a4dafcbcbd7d57a69ed3bc59760381232c2be9c/apps/internal/oauth/ops/authority/authority.go#L146
+type PoPAuthenticationScheme struct {
+	pop.PoPAuthenticationScheme
+}
+
+func (as *PoPAuthenticationScheme) FormatAccessToken(accessToken string) (string, error) {
```
julienstroheker, line 14: Add Uts for all this file, should be easy with a mock token

pkg/internal/token/serviceprincipaltokensecret.go
```diff
@@ -70,6 +70,7 @@ func (p *servicePrincipalToken) getPoPTokenWithClientSecret(
 		p.clientID,
 		p.tenantID,
 		options,
+		nil,
```
julienstroheker, line 73: The more I think about this, I think we should not nil this and make it explicit by instantiating `GetSwPoPKey` inside this package

pkg/pop/poptoken.go
```diff
@@ -0,0 +1,11 @@
+package pop
+
+import "github.com/Azure/kubelogin/pkg/internal/pop"
+
+func GetSwPoPKey() (*SwKey, error) {
```
julienstroheker, line 5: Let's get rid of this and let the caller to instantiate this.

pkg/pop/msal_confidential.go
```diff
@@ -0,0 +1,32 @@
...
+	scopes []string,
+	cred confidential.Credential,
+	authority,
+	clientID,
+	tenantID string,
+	options *azcore.ClientOptions,
+	popKeyFunc func() (*SwKey, error),
+) (string, int64, error) {
+
+	internalPopKeyFunc := func() (*pop.SwKey, error) {
```
julienstroheker, line 23: let's get rid of this and just pass the `popKeyFunc` directly to `AcquirePoPTokenConfidential`

pkg/internal/token/serviceprincipaltokensecret.go
julienstroheker: actually you should be able to invoke `GetSwPoPKey` from pop internal package directly no ?

pkg/pop/poptoken.go
JorgeDaboub: I think the issue is that the caller does not have access to the internal *SwKey object. I tried, but I could only make it work using this function to convert it from the *pop.SwKey.

JorgeDaboub:
Hi @bcho 

This is the follow-up PR to [https://github.com/Azure/kubelogin/pull/447](https://github.com/Azure/kubelogin/pull/447)
As requested, I’ve only exposed the necessary methods. Julien and I have already reviewed the changes.
Could you please take a look? Thanks!

pkg/pop/types.go
```diff
@@ -0,0 +1,30 @@
...
+	pop.PoPAuthenticationScheme
+}
+
+// FormatAccessToken takes an access token, formats it as a PoP token,
+// and returns it as a base-64 encoded string
+func (as *PoPAuthenticationScheme) FormatAccessToken(accessToken string) (string, error) {
+	return as.PoPAuthenticationScheme.FormatAccessToken(accessToken)
+}
+
+type SwKey struct {
```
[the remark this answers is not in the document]
bcho: with this type alias, we can save the extra conversion wrapper in here https://github.com/Azure/kubelogin/pull/542/files#diff-0c11dd5457aab938d78253ff2862d10a8df78316b01fea47eb610cb056cf70cdR25 too

bcho (approved):
LGTM

bcho:
@weinong PTAL

julienstroheker (approved):

julienstroheker:
@rharpavat PTAL

rharpavat (approved):
LGTM!

weinong:
LGTM, but how will these exposed functions be used?

PR #590, opened 2024-12-30
use bingo to manage golangci-lint

weinong:
* use bingo to manage golangci-lint
* bump golangci-lint to 1.62.2
* bump go version to latest N-1 (1.22.10) supported version


go.mod
```diff
@@ -1,7 +1,7 @@
 module github.com/Azure/kubelogin
 
 // NOTE: kubelogin follows the same support policy as Go, which supports the last two major versions.
-go 1.21.11
+go 1.23.4
```
bcho, line 4: this is unexpected? Or we should bump this to 1.22.x to match the support policy comment above

bcho (approved):
LGTM

Issue #555, opened 2024-11-20
kubelogin(interactive pop) issue with AKSArc connected to Winfield


Aijing2333:
@julienstroheker 
For Winfield, we need to disable instance discovery in the auth flow, or we will hit below error. Maybe the previous flow work as the instance discovery was not enabled at that time

PS C:\Users\cloudtest> $env:AZURE_ENVIRONMENT_FILEPATH = "C:\AksArc\env.json"
PS C:\Users\cloudtest> kubectl --kubeconfig C:\Aksarc\config-aksarc-test1-user get ns
Error: failed to get token: failed to create PoP token using interactive login: failed to create PoP token with interactive flow: http call(https://login.microsoftonline.com/common/discovery/instance?api-version=1.1&authorization_endpoint=https%3A%2F%2Flogin.autonomous.cloud.private%2F98b8267d-e97f-426e-8b3f-7956511fd63f%2Foauth2%2Fv2.0%2Fauthorize)(GET) error: reply status code was 400:
{"error":"invalid_instance","error_description":"AADSTS50049: Unknown or invalid instance. Trace ID: 8475d55f-87de-42e3-a4fb-8bc1e1401100 Correlation ID: 7e9eef1b-bbc1-4d55-b2b6-9c3b1ae15fb7 Timestamp: 2024-11-20 03:14:35Z","error_codes":[50049],"timestamp":"2024-11-20 03:14:35Z","trace_id":"8475d55f-87de-42e3-a4fb-8bc1e1401100","correlation_id":"7e9eef1b-bbc1-4d55-b2b6-9c3b1ae15fb7","error_uri":"https://login.microsoftonline.com/error?code=50049"}
E1120 03:14:35.121603    2420 memcache.go:265] couldn't get current server API group list: Get "https://192.168.1.202:6443/api?timeout=32s": getting credentials: exec: executable kubelogin failed with exit code 1
Error: failed to get token: failed to create PoP token using interactive login: failed to create PoP token with interactive flow: http call(https://login.microsoftonline.com/common/discovery/instance?api-version=1.1&authorization_endpoint=https%3A%2F%2Flogin.autonomous.cloud.private%2F98b8267d-e97f-426e-8b3f-7956511fd63f%2Foauth2%2Fv2.0%2Fauthorize)(GET) error: reply status code was 400:

env.json:
{
    "name":  "AzureEdgeCloud",
    "managementPortalURL":  "https://portal.autonomous.cloud.private/",
[...]

PR #593, opened 2025-01-13
Add disable-instance-discovery option in interactive pop mode

Aijing2333:
For Winfield, we need to disable instance discovery in the auth flow, or we will hit error. See details in https://github.com/Azure/kubelogin/issues/555 

Context
Project Winfield is an appliance Virtual Machine (VM) that provides a local control plane for your datacenter. It allows you to deploy and manage the appliance easily and use Azure Arc and Azure management capabilities in a disconnected or air-gapped environment. Project Winfield is designed to help you meet specific regulatory requirements when you need a local control plane.

### DIFF
--- a/docs/book/src/cli/get-token.md
+++ b/docs/book/src/cli/get-token.md
@@ -20,6 +20,7 @@ ZURE_CLIENT_CERTIFICATE_PASSWORD environment variable
       --client-id string                     AAD client application ID. It may be specified in AAD_SERVICE_PRINCIPAL_CLIENT_ID or AZURE_CLIENT_ID environment variable
       --client-secret string                 AAD client application secret. Used in spn login. It may be specified in AAD_SERVICE_PRINCIPAL_CLIENT_SECRET or AZURE_CLIENT_S
 ECRET environment variable
+      --disable-instance-discovery           set to true to disable instance discovery in environments with their own Identity Provider (not Entra ID/AAD) that does not have instance metadata discovery endpoint.
   -e, --environment string                   Azure environment name (default "AzurePublicCloud")
       --federated-token-file string          Workload Identity federated token file. It may be specified in AZURE_FEDERATED_TOKEN_FILE environment variable
   -h, --help                                 help for get-token
--- a/pkg/internal/pop/msal_public.go
+++ b/pkg/internal/pop/msal_public.go
@@ -9,19 +9,24 @@ import (
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/public"
 )
 
+type PublicClientOptions struct {
+	Authority                string
+	ClientID                 string
+	DisableInstanceDiscovery bool
+	Options                  *azcore.ClientOptions
+}
+
 // AcquirePoPTokenInteractive acquires a PoP token using MSAL's interactive login flow.
 // Requires user to authenticate via browser
 func AcquirePoPTokenInteractive(
 	context context.Context,
 	popClaims map[string]string,
 	scopes []string,
-	authority,
-	clientID string,
-	options *azcore.ClientOptions,
+	pcOptions *PublicClientOptions,
 ) (string, int64, error) {
 	var client *public.Client
 	var err error
-	client, err = getPublicClient(authority, clientID, options)
+	client, err = getPublicClient(pcOptions)
 	if err != nil {
 		return "", -1, err
 	}
@@ -53,13 +58,11 @@ func AcquirePoPTokenByUsernamePassword(
 	context context.Context,
 	popClaims map[string]string,
 	scopes []string,
-	authority,
-	clientID,
 	username,
 	password string,
-	options *azcore.ClientOptions,
+	pcOptions *PublicClientOptions,
 ) (string, int64, error) {
-	client, err := getPublicClient(authority, clientID, options)
+	client, err := getPublicClient(pcOptions)
 	if err != nil {
 		return "", -1, err
 	}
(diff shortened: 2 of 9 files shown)


pkg/internal/pop/msal_public.go
```diff
@@ -18,10 +18,11 @@ func AcquirePoPTokenInteractive(
 	authority,
 	clientID string,
 	options *azcore.ClientOptions,
+	isWinfieldEnv bool,
```
julienstroheker, line 21: This parameter is too specific to winfield here, this should be more generic such as `WithInstanceDiscovery` for example.

I also think that we should start passing optional option or struct here.

pkg/internal/token/provider.go
```diff
@@ -34,14 +35,14 @@ func NewTokenProvider(o *Options) (TokenProvider, error) {
 	case DeviceCodeLogin:
 		return newDeviceCodeTokenProvider(*oAuthConfig, o.ClientID, o.ServerID, o.TenantID)
 	case InteractiveLogin:
-		return newInteractiveTokenProvider(*oAuthConfig, o.ClientID, o.ServerID, o.TenantID, popClaimsMap)
+		return newInteractiveTokenProvider(*oAuthConfig, o.ClientID, o.ServerID, o.TenantID, popClaimsMap, isWinfieldEnvironment(o.Environment))
```
julienstroheker, line 38: overall I would like to avoid naming winfield here, and find a broader approach.

pkg/internal/token/provider.go
```diff
@@ -87,3 +88,10 @@ func getAzureEnvironment(environment string) (azure.Environment, error) {
 	}
 	return azure.EnvironmentFromName(environment)
 }
+
+func isPrivateEnvironment(environment string) bool {
```
bganapa, line 92: isAirGappedCloud is likely better naming

pkg/internal/token/options.go
```diff
@@ -38,6 +38,7 @@ type Options struct {
 
 const (
 	defaultEnvironmentName = "AzurePublicCloud"
+	privateEnvironmentName = "AzureStackCloud"
```
bganapa, line 41: who sets the value AzureStackCloud? where is this being consumed?

weinong: I suppose AKS should also set this argument in AGC environment?

weinong: Also, I don't think `AzureStackCloud` is appropriate for Airgapped environment.

Aijing2333: This name is hardcode in https://github.com/Azure/go-autorest/blob/main/autorest/azure/environments.go#L293  to load env from file(retrieve ActiveDirectoryEndpoint), so we set this name.
For AKS in AGC environment, not very familiar with it, not sure if it needs to load own ActiveDirectoryEndpoint and disable discovery instance

weinong: I think the most appropriate way is to add a flag in cli to disable instance metadata discovery since AGC actually has their own instance metadata discovery endpoint

bganapa: thinking back,  AirGappedCloud is a bit overloaded. The proposed change here is specifically for the environment Azure Local Disconnected (https://learn.microsoft.com/en-us/azure/azure-local/manage/disconnected-operations-overview)

It is also referred internally as Arc Autonomous. This disconnected environment has its own simple Identity Provider(not AAD) which does not have instance metadata discovery endpoint. 

@Aijing2333 Is it passible to parameterize DisableInstanceDiscovery  to kubelogin.exe ? who will be setting that option?

weinong: for now, exposing this parameter directly in cli and api and let the end user decide is most appropriate, rather than deriving it from the environment

Aijing2333: Thanks @weinong  @bganapa  for your suggestion. I will make the change to expose the DisableInstanceDiscovery option, but the name AzureStackCloud is still needed as it is the only way to load env from file  https://github.com/Azure/go-autorest/blob/main/autorest/azure/environments.go#L293, is this fine?

bganapa: Do we need to refer AzureStackCloud within Kubelogin codebase or outside of it ? I know we have this reference of AzureStackCloud  in many codebases, CLI, Custom Location RP  etc..  If we cant avoid it, I think we will have to make that case to weinong and take it forward..

Aijing2333: Never mind, I suddenly realize after I expose the DisableInstanceDiscovery option, this code change removed. Kubelogin already contains the logic to the load env file when env name is azurestackcloud

pkg/internal/pop/msal_public.go
```diff
@@ -9,19 +9,24 @@ import (
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/public"
 )
 
+type PublicClientOptions struct {
+	Authority                string
+	ClientID                 string
+	DisableInstanceDiscovery bool
```
bganapa, line 15: Dont we also need TenantId here? wondering as I introduced MsalClientOptions in my other PR and the possibility of reusing it here..

docs/book/src/cli/get-token.md
```diff
@@ -20,6 +20,7 @@ ZURE_CLIENT_CERTIFICATE_PASSWORD environment variable
       --client-id string                     AAD client application ID. It may be specified in AAD_SERVICE_PRINCIPAL_CLIENT_ID or AZURE_CLIENT_ID environment variable
       --client-secret string                 AAD client application secret. Used in spn login. It may be specified in AAD_SERVICE_PRINCIPAL_CLIENT_SECRET or AZURE_CLIENT_S
 ECRET environment variable
+      --disable-instance-discovery           set to true to disable instance discovery in environments with their own simple Identity Provider (not AAD) that do not have instance metadata discovery endpoint.
```
weinong, line 23: what do you mean by `simple`?

weinong, line 23:
```suggestion
      --disable-instance-discovery           set to true to disable instance discovery in environments with their own Identity Provider (not Entra ID/AAD) that does not have instance metadata discovery endpoint.
```

pkg/internal/pop/msal_public.go
weinong: +1. why don't we have https://pkg.go.dev/github.com/AzureAD/microsoft-authentication-library-for-go@v1.3.2/apps/public#WithTenantID

Aijing2333: I think the code uses https://pkg.go.dev/github.com/AzureAD/microsoft-authentication-library-for-go@v1.3.2/apps/public#WithAuthenticationScheme for proof of possession access tokens. WithTenantID can't be used for pop access token. @rharpavat please correct me if any mistake

weinong: You should specify tenant ID by using https://pkg.go.dev/github.com/AzureAD/microsoft-authentication-library-for-go@v1.3.2/apps/public#WithTenantID

pkg/internal/pop/msal_public.go
```diff
@@ -88,23 +91,25 @@ func AcquirePoPTokenByUsernamePassword(
 }
 
 // getPublicClient returns an instance of the msal `public` client based on the provided options
-func getPublicClient(
-	authority,
-	clientID string,
-	options *azcore.ClientOptions,
-) (*public.Client, error) {
+// The instance discovery will be disable on private cloud
```
weinong, line 94:
```suggestion
// The instance discovery should be disabled on private cloud
```

pkg/internal/pop/msal_public.go
Aijing2333: Updated with fix

pkg/internal/pop/msal_public.go
Aijing2333: Fixed

pkg/internal/pop/msal_public.go
bganapa: If we are going to add TenantId here, i think it can be refactored to reuse the MsalClientOptions defined in the msal_confidential.go

Aijing2333: Thanks for the suggestion, reuse MsalClientOptions

weinong (approved):
LGTM

Issue #196, opened 2023-02-14
Empty client secret when defining it from environment variables


alk-adong:
Hello,

We are using a service principal with kubelogin and recently migrated to the following commandline:

```
# Before
kubelogin convert-kubeconfig --login spn --client-id ${ARM_CLIENT_ID} --client-secret ${ARM_CLIENT_SECRET}

# After
kubelogin convert-kubeconfig --login spn --use-azurerm-env-vars
```

However, with the new commandline, in the `~/.kube/config` file, we are missing the `--client-secret` argument and its value.

```
- name: clusterUser_REDACTED
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      args:
      - get-token
      - --login
      - spn
      - --server-id
      - REDACTED
      - --client-id
      - REDACTED
      - --tenant-id
      - REDACTED
      - --environment
      - AzurePublicCloud
      command: kubelogin
      env: null
      provideClusterInfo: false
```

It results in the following error:

```
Error: both clientSecret and clientcert cannot be empty
```

I noticed that changing to the following commandline fixes the issue:

```
kubelogin convert-kubeconfig --login spn --use-azurerm-env-vars --client-secret ""
```

Most likely, this is the responsible part, as it checks that the flag is defined in the commandline arguments instead of checking if the value in the struct is non-empty.

https://github.com/Azure/kubelogin/blob/992e74d0fdb4426b36d49a0486be68b9f230d487/pkg/converter/convert.go#L251-L261

sam-cogan:
It seems there is the same issue with Client ID as well, it's not written out unless you supply an empty value as a parameter

markgould:
This is still an issue. Thankfully this thread helped me resolve it.

PR #594, opened 2025-01-14
Add disable environment override option.

dpersson:
Sometimes you don't want to overwrite the provided arguments with possible settings from your environment variables.
This adds a flag to convert-kubeconfig for disabling environment override. It uses the flag in "kubelogin get-token" UpdateFromEnv() to return early and disregard possible setting of environment variables.

`kubelogin convert-kubeconfig -l spn --client-id xxx --client-secret xxx --disable-environment-override`

will result in:
```

{
  "name": "clusterxxx",
  "user": {
    "exec": {
      "apiVersion": "client.authentication.k8s.io/v1beta1",
      "args": [
        "get-token",
        "--login",
        "spn",
        "--server-id",
        "xxx",
        "--client-id",
        "***",
        "--tenant-id",
        "***",
        "--environment",
        "AzurePublicCloud",
        "--client-secret",
        "xxx",
        "--disable-environment-override"
      ],
      "command": "kubelogin",
      "env": null,
      "installHint": "\nkubelogin is not installed which is required to connect to AAD enabled cluster.\n\nTo learn more, please go to https://aka.ms/aks/kubelogin\n",
      "provideClusterInfo": false
    }
  }
}

```

Provides solution for #196


dpersson:
> @microsoft-github-policy-service agree

@microsoft-github-policy-service agree

weinong (approved):
LGTM

PR #595, opened 2025-01-16
Adding an option to disable instance discovery in AcquirePoPTokenConfidential

bganapa:
In autonomous (completely disconnected) environments to disable authority validation and instance discovery, msal provides an option with WithInstanceDiscovery(false) in the client creation.

The PR introduces a param instanceDiscovery in AcquirePoPTokenConfidential to disable instance discovery and calls WithInstanceDiscovery 

The other code paths where AcquirePoPTokenConfidential is called , the default value of true is set. When AcquirePoPTokenConfidential  is called from Client Side Proxy (arcproxy) , the value will be customized depending on whether the current cloud is autonomous (disconnected) or not

### DIFF
--- a/pkg/internal/token/serviceprincipaltokencertificate.go
+++ b/pkg/internal/token/serviceprincipaltokencertificate.go
@@ -90,6 +90,7 @@ func (p *servicePrincipalToken) getPoPTokenWithClientCert(
 		p.cloud.ActiveDirectoryAuthorityHost,
 		p.clientID,
 		p.tenantID,
+		true,
 		options,
 		pop.GetSwPoPKey,
 	)
--- a/pkg/internal/token/serviceprincipaltokencertificate.go
+++ b/pkg/internal/token/serviceprincipaltokencertificate.go
@@ -87,11 +87,13 @@ func (p *servicePrincipalToken) getPoPTokenWithClientCert(
 		p.popClaims,
 		scopes,
 		cred,
-		p.cloud.ActiveDirectoryAuthorityHost,
-		p.clientID,
-		p.tenantID,
-		true,
-		options,
+		&pop.MsalClientOptions{
+			Authority:                p.cloud.ActiveDirectoryAuthorityHost,
+			ClientID:                 p.clientID,
+			TenantID:                 p.tenantID,
+			DisableInstanceDiscovery: false,
+			Options:                  options,
+		},
 		pop.GetSwPoPKey,
 	)
 	if err != nil {
--- a/pkg/internal/token/serviceprincipaltokencertificate.go
+++ b/pkg/internal/token/serviceprincipaltokencertificate.go
@@ -91,7 +91,7 @@ func (p *servicePrincipalToken) getPoPTokenWithClientCert(
 			Authority:                p.cloud.ActiveDirectoryAuthorityHost,
 			ClientID:                 p.clientID,
 			TenantID:                 p.tenantID,
-			DisableInstanceDiscovery: false,
+			DisableInstanceDiscovery: p.disableInstanceDiscovery,
 			Options:                  options,
 		},
 		pop.GetSwPoPKey,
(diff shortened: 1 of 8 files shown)


JorgeDaboub (approved):

pkg/internal/token/serviceprincipaltokencertificate.go
```diff
@@ -90,6 +90,7 @@ func (p *servicePrincipalToken) getPoPTokenWithClientCert(
 		p.cloud.ActiveDirectoryAuthorityHost,
 		p.clientID,
 		p.tenantID,
```
bganapa, line 92: what is the use case of this method getPoPTokenWithClientCert? This is not being used by CSP, however wondering whether we need to expose InstanceDiscovery  option here as well..

JorgeDaboub:
@bganapa There appears to already be an active PR for Winfield that I was unaware of: https://github.com/Azure/kubelogin/pull/593

bganapa:
> @bganapa There appears to already be an active PR for Winfield that I was unaware of: #593
 I looked at the other PR, two different set of files. I think we can follow a similar approach in both the PRs. @julienstroheker to provide inputs here

JorgeDaboub (approved):

JorgeDaboub:
LGTM

pkg/internal/token/serviceprincipaltokencertificate.go
```diff
@@ -87,10 +87,13 @@ func (p *servicePrincipalToken) getPoPTokenWithClientCert(
...
 		cred,
-		p.cloud.ActiveDirectoryAuthorityHost,
-		p.clientID,
-		p.tenantID,
-		options,
+		&pop.MsalClientOptions{
+			Authority:                p.cloud.ActiveDirectoryAuthorityHost,
+			ClientID:                 p.clientID,
+			TenantID:                 p.tenantID,
+			DisableInstanceDiscovery: false,
```
weinong, line 94: shouldn't this be based on the environment like how it is done in #593 ?

```go
func isAirGappedCloud(environment string) bool {
	if environment == "" {
		return false
	}
	return strings.EqualFold(environment, privateEnvironmentName)
}
```

JorgeDaboub: In this case, don't we want to keep it specific to the feature? There might be other use cases or scenarios besides air-gapped cloud environments that require instance discovery to be off.

Aijing2333: AFAK, AKSArc only supports interactive pop token scenario, so this code path will not used by AKSArc scenario

bganapa: I prefer to leave it as such without introducing the isAirGappedCloud constraint.

weinong: IMO, we should not hardcode it here. Can we add this flag to `servicePrincipalToken` and set it in `func newServicePrincipalTokenProvider()`?

JorgeDaboub: @weinong 
Should we pass the false value into the function [here](https://github.com/Azure/kubelogin/blob/2cd42fbb1a4d98d6b8504cf5fae9d3fd721f1bd4/pkg/internal/token/provider.go#L42C3-L43C1), or set it to default to false inside the function?



`return newServicePrincipalTokenProvider(cloudConfiguration, o.ClientID, o.ClientSecret, o.ClientCert, o.ClientCertPassword, o.ServerID, o.TenantID, popClaimsMap)`

weinong: pass the false value into the function [here](https://github.com/Azure/kubelogin/blob/2cd42fbb1a4d98d6b8504cf5fae9d3fd721f1bd4/pkg/internal/token/provider.go#L42C3-L43C1),

bganapa: making that change now

bganapa: addressed that @weinong

weinong (approved):
LGTM

PR #608, opened 2025-02-23
rewrote token implementation and added official cache support

weinong:
Changes:
1. removed adal from all interfaces and replaced with azidentity
2. rewrote most credential flows using azidentity
3. enabled persistent cache for supported credential flows. For details https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/TOKEN_CACHING.MD
4. every combination of credential flow is written in its own struct
5. added support to use PEM files for client certs
6. added http client to allow recording and playback and go-vcr based unit tests
7. migrated go-vcr to v4
8. added disabling instance discovery option to all azidentiy credentials
9. fixed a bug where authority host is not set correct in some environments

Breaking change:
1. Previous caching implementation is removed. Now we are using caching provided by azidentity. This also means any credential flows not implemented by azidentity will not have any caching. Notably, interactive with pop, device code with legacy and ropc with pop will NOT have cache.
2. The binary is now built with CGO enabled to allow secure token caching on the host


pkg/internal/token/options.go
```diff
@@ -229,6 +252,30 @@ func (o *Options) UpdateFromEnv() {
 	}
 }
 
+func (o *Options) GetCloudConfiguration() cloud.Configuration {
+	if o.AuthorityHost != "" {
```
Aijing2333, line 256: For Airgapped environment, the way to set the AuthorityHost parameter is directly in cli or api instead of loading from environment(with name AzureStackCloud), right?

Aijing2333 (approved):

pkg/internal/token/azuredevopscredential.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+	return &AzureDeveloperCLICredential{cred: cred}, nil
+}
+
+func (c *AzureDeveloperCLICredential) Name() string {
+	return "AzureDeveloperCLICredential"
+}
+
+func (c *AzureDeveloperCLICredential) Authenticate(ctx context.Context, opts *policy.TokenRequestOptions) (azidentity.AuthenticationRecord, error) {
+	panic("not implemented")
```
bcho, line 36: should we just return an error in this case instead of panic?

pkg/internal/token/clientcertcredential.go
```diff
@@ -0,0 +1,206 @@
...
+		return nil, fmt.Errorf("client certificate cannot be empty")
+	}
+	var (
+		c   azidentity.Cache
+		err error
+	)
+	if opts.UsePersistentCache {
+		c, err = cache.New(nil)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create cache: %s", err)
```
bcho, line 42: nit: use %w to wrapp error

pkg/internal/token/clientcertcredential.go
```diff
@@ -0,0 +1,206 @@
...
+			break
+		}
+
+		cert, err = x509.ParseCertificate(certBlock.Bytes)
+		if err != nil {
+			return nil, nil, fmt.Errorf("unable to parse certificate. %w", err)
+		}
+
+		certPublicKey, ok := cert.PublicKey.(*rsa.PublicKey)
+		if ok {
```
bcho, line 160: nit: how about we combine the nested if checks together?

```
if ok && isPublicKeyEqual(certPublicKey, &privateKey.PublicKey) {
  found = true
  break
}
```

pkg/internal/token/options.go
weinong: correct

pkg/internal/token/adalclientcertcredential_test.go
```diff
@@ -0,0 +1,99 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNewADALClientCertCredential(t *testing.T) {
+	testCases := []struct {
+		name           string
+		opts           *Options
+		expectError    bool
```
stevekuznetsov, line 13: `expectError == (expectErrorMsg != "")` and `expectNil == (expectName == "")` - not sure you want your test case to allow e.g. 

```
{
    expectError: false,
    expectErrorMsg: "oops",
    expectNil: true,
    expectName: "oops!!",
}
```

pkg/internal/token/adalclientcertcredential_test.go
```diff
@@ -0,0 +1,99 @@
...
+			opts: &Options{
+				ClientID:           "test-client-id",
+				TenantID:           "test-tenant-id",
+				ClientCert:         "test-cert-path",
+				ClientCertPassword: "test-cert-password",
+				IsLegacy:           true,
+			},
+			expectError: false,
+			expectNil:   false,
+			expectName:  "ADALClientCertCredential",
```
stevekuznetsov, line 29: What's the utility of testing this?

```go
func (c *ADALClientCertCredential) Name() string {
	return "ADALClientCertCredential"
}
```

stevekuznetsov: (and not testing the fields in the returned `ADALClientCertCredential`, for instance? To whatever level we even want to test a constructor like `newADALClientCertCredential`, this set of tests allows an implementation that validates input options but doesn't actually create any output to pass, which seems like the only thing we would want to test ;) )

pkg/internal/token/clientcertcredential_test.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	token, err := cred.GetToken(context.Background(), policy.TokenRequestOptions{
+		Scopes: []string{opts.ServerID + "/.default"},
+	})
+	assert.NoError(t, err)
+	assert.Equal(t, testutils.TestToken, token.Token)
+}
+
+func createSelfSignedCertificatePEM() (string, string, error) {
```
stevekuznetsov, line 62: In my experience, if you want to actually assert in the happy case, you need to a) allow the user to seed your source of randomness and b) inject a fake clock for `time.Now()`. The happy-case test only asserts that the returned type outputs the correct value for `.Name()`, which is of dubious value IMO as a test. See [here](https://github.com/openshift/hypershift/blob/4877d3623fa808cb161e6db4d3e1179f08066a43/control-plane-pki-operator/certificaterevocationcontroller/certificaterevocationcontroller_test.go#L61) for an implementation of fake PKI which allows for `cmp.Diff()` style asserts over the output, while checking in generated PKI at development time so as to not tax CI systems' source of randomness. On highly multi-tenant systems for CI I've seen randomness exhaustion cause test flakiness and slow runtime.

pkg/internal/token/clientcertcredential.go
bcho: no, we only have golangci-lint enabled at this moment. We should enable go vet etc

pkg/internal/token/clientcertcredential_test.go
weinong: actually, the intent is just to have a pem file to feed to sdk. it serves nothing else. the http recording only validates the url, verb, and form values. so this pki can be anything

pkg/internal/token/clientcertcredential.go
weinong: i can enable it in next PR

pkg/internal/token/clientcertcredential.go
weinong: fixed in next commit

pkg/internal/token/adalclientcertcredential_test.go
weinong: apologize that you have to debug for Copilot :D

pkg/internal/token/adalclientcertcredential_test.go
weinong: yeah, these tests are just for the sake of coverage. most of the credential flows based on track2 have the output/behavior validation. Not this one, though. Too lazy to do this for adal which is not supported anymore :P

pkg/internal/token/clientcertcredential_test.go
stevekuznetsov: ACK. Still probably good to commit them to disk to reduce load on randomness in CI

weinong: ah. you mean checking in a pem file. yes, i can do that

weinong: updated

bcho (approved):

PR #662, opened 2025-05-02
Jorgedaboub/pop cache

JorgeDaboub:
### Add caching support for PoP token acquisition

This PR introduces support for token caching in the PoP token acquisition flow using MSAL's confidential client. Key changes include:

- Added `ClientOptions` and `WithCustomCache` functional option to support passing a cache to the confidential client.
- Refactored `NewConfidentialClient` to conditionally apply the cache when provided.
- Updated `AcquirePoPTokenConfidential` to accept a pre-initialized `confidential.Client` instead of creating it inline, allowing reuse and cache support.
- Modified test cases and credential construction logic to use the updated `NewConfidentialClient` and pass the client explicitly.

These changes enable improved performance and reduced token acquisition latency by leveraging silent token retrieval from cache when available.

### DIFF
--- a/pkg/internal/pop/msal_confidential.go
+++ b/pkg/internal/pop/msal_confidential.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/cache"
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/confidential"
 )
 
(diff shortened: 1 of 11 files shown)


pkg/internal/pop/msal_confidential.go
```diff
@@ -17,6 +18,70 @@ type MsalClientOptions struct {
...
+// ClientOptions holds options for creating a confidential client
+type ClientOptions struct {
+	Cache cache.ExportReplace
+}
+
+// ConfidentialClientOption defines a functional option for configuring a confidential client
+type ConfidentialClientOption func(*ClientOptions)
+
+// WithCustomCache adds a custom cache to the confidential client
+func WithCustomCache(cache cache.ExportReplace) ConfidentialClientOption {
```
pkoushik, line 30: Is this actually a custom approach or is this done elsewhere in MSAL?

JorgeDaboub: The cache cache.ExportReplace object can be passed to the confidential.New() Method when creating the client. 
https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/b4b8bfc9569042572ccb82b648ea509075fadb74/apps/tests/devapps/client_certificate_sample.go#L41

pkg/internal/pop/msal_confidential.go
```diff
@@ -41,45 +106,19 @@ func AcquirePoPTokenConfidential(
...
-	}
-	if err != nil {
-		return "", -1, fmt.Errorf("unable to create confidential client: %w", err)
-	}
 	result, err := client.AcquireTokenSilent(
 		context,
 		scopes,
 		confidential.WithAuthenticationScheme(authnScheme),
-		confidential.WithTenantID(msalOptions.TenantID),
+		confidential.WithTenantID(tenantID),
```
weinong, line 114: can this line and L121 be done in `NewConfidentialClient()` as well?

JorgeDaboub: We could move those lines into NewConfidentialClient, yes. I left them here because if the client is initialized with AdditionalTenants, we may want to issue tokens for them, this can happen in the CSP flow.

As for the authentication scheme, we could move that too, but doing so would shift the PoP logic into the client initialization. In CSP, we create our own client and then just call a method to acquire the token.

weinong (approved):
LGTM

PR #690, opened 2025-07-26
Update asdf installation instructions

o-farooq:
asdf global has been replaced with asdf set 

https://asdf-vm.com/guide/upgrading-to-v0-16.html#asdf-global-and-asdf-local-commands-have-been-replaced-with-asdf-set


weinong (approved):
LGTM

PR #715, opened 2025-10-14
[PoP token Cache] Fallback to no caching when encrypted cache creation fails 

vineeth-thumma:
Fallback to no caching when encrypted caching initialization fails due to reasons like restricted kernel keyring access, etc. 

- Only initialize PoP cache explicitly when the pop flag is set
- Fallback to not use pop token caching when pop cache initialization fails
- Proactively test storage capability before creating the cache
Reference: https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/cache/cache.go
- Implement PoPKeyProvider interface

Relevant WI: https://msazure.visualstudio.com/AzureArcPlatform/_workitems/edit/35560158

### DIFF
--- a/pkg/internal/token/options.go
+++ b/pkg/internal/token/options.go
@@ -390,7 +390,7 @@ func (o *Options) setPoPTokenCache(cache *popcache.Cache) {
 }
 
 // GetPoPKeyProvider returns a PoPKeyProvider based on the current cache configuration.
-// This centralizes the key provider logic and eliminates code duplication across credential types.
+// This centralizes the key provider logic.
 func (o *Options) GetPoPKeyProvider() PoPKeyProvider {
 	return &defaultPoPKeyProvider{
 		cacheDir: o.getCacheDir(),


pkg/internal/token/execCredentialPlugin.go
```diff
@@ -34,10 +34,17 @@ var errAuthenticateNotSupported = errors.New("authenticate is not supported")
...
 
-	// Create PoP token cache using the official MSAL & MSAL extension libraries.
-	popTokenCache, err := popcache.NewCache(o.AuthRecordCacheDir)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create PoP token cache provider: %w", err)
+	var popTokenCache *popcache.Cache
+	var err error
+	if o.IsPoPTokenEnabled {
+		// Create PoP token cache using the official MSAL & MSAL extension libraries.
+		popTokenCache, err = popcache.NewCache(o.AuthRecordCacheDir)
```
weinong: how about saving the cache pointer in Options as private field? I think that can minimize repetitive changes in all of these login implementations

vineeth-thumma: Updated

pkg/internal/token/clientcertcredentialwithpop.go
```diff
@@ -11,21 +11,21 @@ import (
...
-	cred      confidential.Credential
-	client    confidential.Client
-	options   *pop.MsalClientOptions
-	cacheDir  string
+	popClaims         map[string]string
+	cred              confidential.Credential
+	client            confidential.Client
+	options           *pop.MsalClientOptions
+	cacheDir          string
+	usePersistentKeys bool
```
weinong: the nature of `usePersistentKeys` can be replaced by checking if `cacheDir` is empty or not.

pkg/internal/token/execCredentialPlugin.go
```diff
@@ -24,20 +23,27 @@ type ExecCredentialPlugin interface {
...
 
-	// Create PoP token cache using the official MSAL & MSAL extension libraries.
-	popTokenCache, err := popcache.NewCache(o.AuthRecordCacheDir)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create PoP token cache provider: %w", err)
+	// Initialize PoP token cache in Options if enabled
+	if o.IsPoPTokenEnabled && o.popTokenCache == nil {
+		// Create PoP token cache using the official MSAL & MSAL extension libraries.
+		popTokenCache, err := popcache.NewCache(o.AuthRecordCacheDir)
+		if err != nil {
```
weinong, line 39: I think the logic here can move to options.go such that in the credential's constructor, we just do 
```go
popKeyProvider := o.GetPoPKeyProvider()
```
in GetToken
```go
key := popKeyProvider.GetKey() // sw key or persistent key
```

vineeth-thumma: Good point, updated

weinong (approved):

PR #733, opened 2026-01-12
fix: Multi arch docker build

hilariocoelho:
Docker build was hardcoded for `amd64` only

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,8 @@
 FROM scratch
 
 # Build arguments for multi-architecture support
-ARG TARGETARCH=amd64
+ARG TARGETOS
+ARG TARGETARCH
 ARG VERSION=""
 
 # OpenContainers Image Spec labels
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -9,7 +9,8 @@
 FROM scratch
 
 # Build arguments for multi-architecture support
-ARG TARGETARCH=amd64
+ARG TARGETOS
+ARG TARGETARCH
```
weinong, line 13: can you add default os and arch?

hilariocoelho: What do you mean exactly with "default os and arch"? Those built arguments are already filled on docker buildx by specifying `--platform` argument (ie: `docker build . --platform linux/amd64 --platform linux/arm64`).

When setting `ARG TARGETARCH=amd64`, only `amd64` binary is being included on both containers (`linux/amd64` and `linux/arm64`).

On the github action it is already [prepared](https://github.com/Azure/kubelogin/blob/6e8b5babc994d57e752e24297dd9c6f59247a1f8/.github/workflows/docker-publish.yml#L76) for `linux/amd64` and `linux/arm64`

Issue #734, opened 2026-01-23
PoP token support crashes with nil pointer in cache.Replace when running non-root (v0.2.14)


JS-Jake:
## Summary
kubelogin v0.2.14 crashes with a nil pointer dereference in the PoP token cache when running as a non-root user. The crash occurs in `pkg/internal/pop/cache/cache.go:127` during the `AcquireTokenSilent` call. Running the same command as root succeeds without issue.

## Environment
- **kubelogin version:** v0.2.14
- **Platform:** linux/amd64
- **Base image:** Ubuntu
- **Auth method:** Service Principal (spn) with PoP tokens enabled
- **PoP claims:** `u=/subscriptions/.../connectedClusters/...`

## Reproduction Steps
```bash
# Running as non-root user - FAILS
export AZURE_TENANT_ID="<redacted>"
export AZURE_CLIENT_ID="<redacted>"
export AZURE_CLIENT_SECRET="<secret>"

kubelogin get-token \
  --login spn \
  --server-id 6256c85f-0aad-4d50-b960-e6e9b21efe35 \
  --pop-enabled true \
  --pop-claims "u=/subscriptions/<redacted>/resourceGroups/<redacted>/providers/Microsoft.Kubernetes/connectedClusters/<redacted>"
```
## Error Stack Trace
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xa86c66]

goroutine 1 [running]:
github.com/Azure/kubelogin/pkg/internal/pop/cache.(*Cache).Replace(0x27b0d90?, {0x1b1a768?, 0xc0000c8070?}, {0x7141096d0c60, 0xc0000e2050}, {{0x5?, 0x1?}})
        /home/runner/work/kubelogin/kubelogin/pkg/internal/pop/cache/cache.go:127 +0x26
github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/base.Client.AcquireTokenSilent(...)
        /home/runner/go/pkg/mod/github.com/!azure!a!d/microsoft-authentication-library-for-go@v1.4.2/apps/internal/base/base.go:339 +0x1007
```
## Testing & Findings
- Same command succeeds when running as root
- Reverting to v0.2.8 (before PoP cache implementation) works as non-root without crashing
- PoP tokens generated by v0.2.8 are valid (contain cnf claim with JWK)
- Issue is specific to non-root execution; root execution works
[...]

PR #736, opened 2026-02-09
[Bug Fix] - PoP token flow crash with nil pointer in cache.Replace when running non-root

vineeth-thumma:
Fix bug - > PoP token support crashes with nil pointer in cache.Replace when running non-root

https://github.com/Azure/kubelogin/issues/734

kubelogin v0.2.14 panics with a nil pointer dereference in `cache.(*Cache).Replace` when PoP tokens are enabled and cache creation fails (typically non-root users or container environments without keyring access).

**Root Cause**
A Go nil-interface semantics bug. When NewCache() fails, a typed nil `*popcache.Cache` was stored in Options and later returned by `GetPoPTokenCache()`. When this typed nil was passed to an interface-typed parameter `cache.ExportReplace`, Go wrapped it into a non-nil interface (it carries type info). The downstream nil guard if cache != nil passed, MSAL received the nil-wrapping cache, and called Replace() on a nil receiver → panic.

### DIFF
--- a/pkg/internal/token/options_test.go
+++ b/pkg/internal/token/options_test.go
@@ -681,3 +681,40 @@ func TestAuthorityHostValidation(t *testing.T) {
 		})
 	}
 }
+
+// TestGetPoPTokenCache_NilReturnsUntypedNil verifies that GetPoPTokenCache()
+// returns a true untyped nil (not a typed nil *Cache wrapped in an interface)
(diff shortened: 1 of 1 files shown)


julienstroheker (approved):

weinong (approved):
LGTM

PR #760, opened 2026-06-09
Return errors when cache cleanup fails

pathei-kosmos:
`remove-cache-dir` and deprecated `remove-tokens` previously logged `os.RemoveAll` failures only at verbose level and still returned success. This change returns the removal error with context while preserving success for missing cache paths.

Tests cover both commands with explicit temporary cache paths, including a deterministic removal failure using a regular file as an intermediate path component.


weinong (approved):
LGTM

PR #794, opened 2026-07-21
test: cover NeedAuthenticate for all credential types

NaeemH:
Adds a table-driven test covering `NeedAuthenticate()` across all 18 credential types in `pkg/internal/token`.

Beyond raising coverage, it locks in a behavioral invariant: only the interactive / public-client flows report `true` (they need the MSAL `Authenticate()` step before `GetToken`):

- `true`: `DeviceCodeCredential`, `InteractiveBrowserCredential`, `UsernamePasswordCredential`
- `false`: everyone else (they authenticate as part of `GetToken`)

### Notes
- Uses an inline `interface{ NeedAuthenticate() bool }` so the test depends only on the method under test.
- `NeedAuthenticate()` is a constant predicate on each type, so zero-value instances suffice — no fixtures, no new deps.

### Coverage
The 18 `NeedAuthenticate()` methods go from 0% to covered; package `pkg/internal/token` coverage improves **66.4% -> 68.4%**.

Related to #7.

### DIFF
--- a/pkg/internal/token/needauthenticate_test.go
+++ b/pkg/internal/token/needauthenticate_test.go
@@ -1,0 +1,47 @@
+package token
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// TestNeedAuthenticate locks in which credential types require the MSAL
+// Authenticate() step before GetToken. Only the interactive / public-client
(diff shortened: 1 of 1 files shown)


weinong (approved):
LGTM

